PR #2, opened 2022-05-06
LTI Procotoring Service Specification Proof-Of-Concept: LTI Launch

MichaelRoytman:
**JIRA:** https://openedx.atlassian.net/browse/MST-1260

### Setup Steps

1. Clone the [xblock-lti-consumer](https://github.com/openedx/xblock-lti-consumer) repository into your `{workspace}/src` directory. `{workspace}` should be the directory in which you have your other edX git repositories and should be the same directory your `devstack` repository is located in.
2. Check out the `mroytman/MST-1260-special-exams-LTI-launch-POC` branch locally.
3. In `edx-exams`, create a virtual environment.
4. Activate the virtual environment.
5. Install requirements with `make requirements`.
6. Run `make docker_build` to build the Docker image from the `Dockerfile`.
7. Run `make dev.provision` to provision and start the container.
8. Shell into the container by running `make app-shell`. Once inside the shell, run `pip install -e /edx/src/xblock-lti-consumer`. This will install the branch you checked out locally as an editable installation. Alternatively, you can just avoid the step of checking out the `xblock-lti-consumer` repository if you specify a commit-ish to install.

### Testing Steps

If you want to test this implementation against the IMS Proctoring Specification testing tool, follow these steps.

1. Sign up for an IMS account as an employee of 2U [here](https://site.imsglobal.org/user/register), if you are not already a member. The approval process should take around a day.
2. You will need to set up port forwarding so that your running Docker container can connect with the IMS testing tool. You can use `ngrok` for this or any other tool you like. `ngrok` has comprehensive documentation for how to accomplish this.
3. Once your membership has been approved, go to the testing tool [here](https://certification.imsglobal.org/proctoring/launch).
4. In the testing tool, click `Get Started`.
5. Enter your name, email, and organization and click `Confirm`.
6.  Click `Create` under `Create New Product`.
7. Enter a `Product` and `Version` and click `Confirm`. These values can be anything for the purposes of testing.
8. Click `Select` under `Assessment Platform`.
9. Click `Begin Configuration`.
10. Open up the Django admin by visiting `localhost:8000/admin` and logging in with the default edX development login credentials.
11. Under `LTI_CONSUMER > Lti configuration`, click `Add`.
12. Enter the following configuration:
13.1. `Version`: `LTI 1.3 (with LTI Advantage Support)`
13.2. `Config store`: `Configuration Stored on this model`.
14. From the testing tool, copy the following configuration into the `Lti configuration` model in the Django admin.
14.1. `Tool End-Point to Receive Launches` > `Lti 1p3 launch url`
14.2. `OIDC Login Initiation URL` > `Lti 1p3 oidc url`
15. Click `Next`.
16. Enter the following configuration in the testing tool. In this example, `{root}` refers to the URL that is being forwarded. For example, `https://9c11-73-238-153-173.ngrok.io`.
16.1. `Testing iss Value`: `http://localhost:18000`
16.2. `OIDC Auth URL`: `{root}/api/v1/oidc_auth_request`
16.3. `Platform Well-Known/JWKS URL`: `{root}/api/v1/keyset`
16.4. `Client Id (Assigned to Cert Suite)` > Enter the `Lti 1p3 client id` from the `Lti configuration` model in the Django admin.
17. Click `Next`.
18. Enter `1` for the deployment ID and click `Next`.
19. From the testing tool, copy the`Proctoring Tool JWKS URL` into the `Lti configuration` model in the Django admin in the `Lti 1p3 tool keyset url` field.
20. Click `Next`.
21. Check `Pause on OIDC Exchanges to Inspect Values` and/or `Pause Prior to JWT Messages to Inspect Values` depending on whether you want interstitials before Tool requests or responses to the Assessment Platform.
22. Visit `{root}/api/v1/lti_start_proctoring` to kick off the flow.


Makefile
```diff
@@ -134,32 +138,28 @@ start-devstack: ## run a local development copy of the server
...
+	docker build -t edx_exams:latest -f docker/build/edx-exams/Dockerfile git://github.com/edx/configuration
 
 detect_changed_source_translations: ## check if translation files are up-to-date
 	cd edx_exams && i18n_tool changed
 
 validate_translations: fake_translations detect_changed_source_translations ## install fake translations and check if translation files are up-to-date
 
 docker_build:
-	docker build . -f Dockerfile -t openedx/edx-exams
-	docker build . -f Dockerfile --target newrelic -t openedx/edx-exams:latest-newrelic
```
MichaelRoytman, line 149: There is no `newrelic` target in the Dockerfile. It was removed in this [PR](https://github.com/openedx/edx-cookiecutters/pull/173) from the cookiecutter. I'll plan to remove it from the cookiecutter soon.

Makefile
```diff
@@ -174,8 +174,12 @@ dev.down: # Kills containers and all of their data that isn't in volumes
 dev.stop: # Stops containers so they can be restarted
 	docker-compose stop
 
+dev.provision:
```
MichaelRoytman, line 177: If you run the provision script from your local shell, your environment won't have access to the `DEVSTACK_WORKSPACE` environment variable, defined in `options.mk`. You can export it to your environment, of course, but using this command will be easier and require fewer steps.

schenedx (approved):
👍  I think even if the code here include mocked and fake configuration data, it is valuable to merge as the first step of this repo working with xblock-lti-consumer library. Since the service itself is not production ready, we can be more experimental.

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,287 @@
...
+    # TODO: Here we'd do all the start of proctoring things
+
+    lti_message_hint = 'LtiStartProctoring'
+    preflight_url = get_lti_preflight_url(lti_message_hint)
+
+    return redirect(preflight_url)
+
+def lti_end_assessment(request):
+    """
+    This is the view that'll redirect the learner to the Tool's initial OIDC login initiation URL.
```
ashultz0, line 46: comment copy/pasted?

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,287 @@
...
+    })
+
+    # This template renders an auto-submitting form, which makes a POST request to the redirect_uri, specified in the Tool's response
+    # to the request the Assessment Platform made to the Tool's OIDC login initiation URL.
+    return render(request, 'api/lti_launch_request_form.html', context)
+
+# We do not want Django's CSRF protection enabled for POSTs made by external services to this endpoint.
+# Please see the comment for the lti_oidc_authentication view for a more detailed justification.
+@csrf_exempt
+# Per the Proctoring Services specification, the Tool should only make a POST request to this endpoint.
```
ashultz0, line 192: these comments read like proctorio or whoever is directly calling this view

but really proctorio is telling the learner browser to call this view right? That makes a huge difference in what we expect to see in the request. Although we still wouldn't expect it to be able to provide a CSRF token in a form field

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,287 @@
...
+@require_http_methods(['POST'])
+def lti_start_assessment(request):
+    """
+    This is the view that the Tool will make a request to when the leaner is ready to begin the assessment
+    from the Tool's point of view.
```
ashultz0, line 197: it might be helpful to have a comment at the top of the file with the standard call ordering we expect to see for an exam. And possibly split the view file into views called from edx vs from redirect by the tool OR calls which start vs calls which end the exam. Which is two different possible splits.  Mostly this just keeps the amount to digest in a single file down.

edx_exams/apps/api/api.py
```diff
@@ -0,0 +1,52 @@
+from django.conf import settings
```
ashultz0, line 1: should this be under the already made v1 directory?

edx_exams/apps/api/v1/urls.py
```diff
@@ -1,4 +1,13 @@
 """ API v1 URLs. """
+from django.urls import path
+
+from . import views
 
 app_name = 'v1'
-urlpatterns = []
+urlpatterns = [
+    path('lti_start_proctoring', views.lti_start_proctoring),
+    path('oidc_auth_request', views.lti_oidc_authentication),
```
ashultz0, line 9: if this isn't mandated to be oidc_auth_request in the path it would be clearer to have these all start with lti

or possibly just have them more like a hierarchy all under lti/

Zacharis278:
still browsing through this but if we wanted to merge just so we have the code set could we maybe move this into its own folder/app? That way everything here doesn't get tangled up in the same namespace as we build out the service. There should probably be a separation between any of the LTI urls/apis we need and and the core models/views that the MFE/LMS will interact with anyway.

Zacharis278:
@MichaelRoytman just to check my understanding. Once we get further into modifying the lti-consumer library much of the the api/views.py will become unnecessary correct? Most of this seems to parallel functions that already exist there but just need to be expanded or have some proctoring stuff thrown on top.

MichaelRoytman:
@Zacharis278 I wrote the POC with the current state of the `xblock-lti-consumer` library in mind. I did consider making the necessary modifications to `xblock-lti-consumer` to do what you're suggesting, but I felt it was out of scope based on your discovery on decoupling. I could use your help understanding where you envision the `xblock-lti-consumer` will be after we refactor it. I was absent when you reviewed your discovery document for the decoupling of `xblock-lti-consumer`, and I did review it myself after the fact, but I may be missing something.

I do see views in `lti_consumer/plugin/views.py`, but they're part of a plugin that's plugged into the LMS. It also references the xBlock configuration exclusively. Is it fair to say that the plan is that this plugin will be made to be more generic and will use the `LtiConfiguration` model over the xBlock? It will then serve as a plugin to this service, providing the necessary views for LTI proctoring? Most of the overlapping logic between this code and the `xblock-lti-consumer` is in the actual xBlock handler - that will be pulled out and made generic as well, right?

If so, yes, most of this can be moved into the consumer. We still will need a way to pass necessary data owned by `edx-exams` to the library code (e.g. the resource_link, user identity claims, etc.), but we would do that through the `LtiAdvantageConsumer` class, as we are doing in this POC.

Am I making the right assumptions here?

edx_exams/apps/api/v1/views.py
MichaelRoytman: I see what you mean. Does something like the following sound better to you?

"Per the Proctoring Services specification, the Tool should only direct the learner's browser to make a POST request to this endpoint."

If so, I can comb through the comments and make sure it's clear that the browser is making the requests.

ashultz0: yeah something like that would be fine

edx_exams/apps/api/v1/views.py
MichaelRoytman: Because the endAssessment message is a new Assessment Platform originating message, it has to go through the entire OpenID Connect LTI launch flow all over again in order to send the endAssessment message, so this is technically still a correct comment. I can clarify the comment with that detail. Is this something else unclear about the comment?

ashultz0: so the first part of ending is to login? That is super confusing but "has to go through the entire OpenID Connect LTI launch flow all over again in order to send the endAssessment message" would clear up why we're going back to the same place

Zacharis278:
@MichaelRoytman that all sounds correct. And for a POC, yeah, trying to modify the consumer library would probably be a nightmare to tackle straight off. If all goes well I hope we can modify that library so very little LTI specifics would need to exist here. The biggest two bits are allowing the use of the LtiConfiguration Django model instead of modulestore and moving all of the xblock dependent code into the more generic LTI consumer component.

This helps us, and would basically allow anyone to build an LTI integration in platform that isn't tied to a course unit.

Since this already works I do think it's helpful to have the code in here, we should just put in its own box since it's going to become slowly unused as we progress.

edx_exams/apps/api/v1/views.py
MichaelRoytman: "Login" is kind of confusing here. More specifically, calling this endpoint initiates the OpenID Connect authorization flow, which is a necessary step when sending any LTI related message, because the Assessment Platform acts as the identity provider here. The IMS calls the endpoint that does this the "3rd Party initiated login end-point". The rationale gets more into the OIDC specification, which the IMS pulls heavily from, which I'm not as familiar with. I'll see if I can make this clearer.

ashultz0 (approved):

PR #4, opened 2022-05-17
docs: adr for brokering edx-proctoring calls

Zacharis278:
## [MST-1452](https://2u-internal.atlassian.net/browse/MST-1452)

Based on the discovery work at: https://2u-internal.atlassian.net/wiki/spaces/PT/pages/15438674/Discovery+Support+multiple+proctoring+subsystems

There's a list of each API call we currently make [here](https://2u-internal.atlassian.net/wiki/spaces/PT/pages/15439580/Technical+Reference+--+State+of+Proctoring#edx-proctoring-python-API). Although, by this solution edx-exams would only end up handling a subset of these.

### DIFF
--- a/docs/decisions/0002-multiple-exam-systems.rst
+++ b/docs/decisions/0002-multiple-exam-systems.rst
@@ -1,0 +1,56 @@
+Support for Multiple Exam Systems
+=================================
+
+Status
+------
+
+Propsed (circa May 2022)
+
+Context
+-------
+The addition of this service to Open edX will result in two separate services
+for managing exams, this IDA and the `edx-proctoring`_ plugin. Because this IDA will
+not support existing vendor integrations with the same feature set, edx-proctoring cannot be deprecated
+right away. As a result, we need a solution that allows course teams to choose a proctoring tool
+that is set up in either service but avoids adding configuration of lots of switching code to edx-platform.
+
+Decision
+--------
+This service will include all logic necessary to determine which proctoring implementation a particular API
+request is relevant for and forward requests `edx-proctoring`_ where appropriate.
+
+#. edx-platform will not have configuration or state indicating if an exam or proctoring provider is configured
+   by edx-exams or edx-proctoring.
+
+#. Any request for an API function that exists in this codebase will call the edx-exams REST API only.
+
+    * edx-exams will determine if a request should be forwarded to edx-proctoring or be processed internally.
+
+#. Any request for an API function that does not have an implementation in edx-exams will
+   continue to use the edx-proctoring python api.
+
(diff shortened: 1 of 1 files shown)


docs/decisions/0002-multiple-exam-systems.rst
```diff
@@ -0,0 +1,56 @@
...
+The addition of this service to Open edX will result in two separate services
+for managing exams, this IDA and the `edx-proctoring`_ plugin. Because this IDA will
+not support existing vendor integrations with the same feature set, edx-proctoring cannot be deprecated
+right away. As a result, we need a solution that allows course teams to choose a proctoring tool
+that is set up in either service but avoids adding configuration of lots of switching code to edx-platform.
+
+Decision
+--------
+This service will include all logic necessary to determine which proctoring implementation a particular API
+request is relevant for and forward requests `edx-proctoring`_ where appropriate.
```
ashultz0, line 20: requests to 'edx-proctoring'

docs/decisions/0002-multiple-exam-systems.rst
```diff
@@ -0,0 +1,56 @@
...
+
+    * edx-exams will determine if a request should be forwarded to edx-proctoring or be processed internally.
+
+#. Any request for an API function that does not have an implementation in edx-exams will
+   continue to use the edx-proctoring python api.
+
+    * These features should disable gracefully if no configuration is found.
+
+
+Consesqunces
```
ashultz0, line 35: consequences

docs/decisions/0002-multiple-exam-systems.rst
ashultz0: Also terminology nit - forward also is a bit of a weird word because it sounds like it's going to another process somewhere, but we'll just have edx-proctoring lib embedded in the exam service right?

maybe "delegate" rather than "forward" ?

Zacharis278: well we can't actually embed it b/c that would involve pulling the existing database over and rebuilding the entire edx-proctoring python API with rest endpoints. The proposal here would only require rest endpoints for the intersection of calls both edx-exams and edx-proctoring need.

ashultz0: ahhhh I see carry on :)

ashultz0 (approved):

Zacharis278:
@schenedx would like to get your thoughts since it's the implementation of your discovery work.

schenedx (approved):
👍  All make sense here. Thank you for the summary and the decision doc.

PR #6, opened 2022-05-23
feat: add local provision script

alangsto:
Updated paths for dockerfile, removed unnecessary database file (mistakenly added as part of initial cookie cutter), and add additional provision script for local set up. 

The local provision script mirrors what commerce-coordinator has done, as neither commerce-coordinator nor edx-exams have been added to devstack yet.


schenedx (approved):
👍

PR #8, opened 2022-05-27
feat: add API docs

alangsto:
## [MST-1445](https://2u-internal.atlassian.net/browse/MST-1445)

Add api docs for edx-exams. In order for the health endpoint to be listed, I had to convert it to an API View. This might be a little hacky, so we could leave it as is and just have the API docs be empty for now. 

Screen Shot 2022-05-27 at 11 29 05 AM

Screen Shot 2022-05-27 at 11 29 12 AM


schenedx (approved):
👍

PR #10, opened 2022-06-06
feat: add exam models

alangsto:
Updated the User model, and added

- Exam
- ExamAttempt
- ProctoringProvider
- CourseExamConfiguration

Still need to add the migration, but will do that after getting review that the fields look appropriate.


edx_exams/apps/core/models.py
```diff
@@ -16,6 +16,12 @@ class User(AbstractUser):
     """
     full_name = models.CharField(_('Full Name'), max_length=255, blank=True, null=True)
 
+    first_name = models.CharField(max_length=255, blank=True, null=True)
```
alangsto, line 19: Not sure if first name and last name are needed if full name is already part of the model, @Zacharis278 is there a reason we would want to separate the name?

ashultz0: +1 on leaving them out, not least because "first name / last name" is not a universal concept anyway

edx_exams/apps/api/models.py
```diff
@@ -1,4 +1,108 @@
-# Models that can be shared across multiple versions of the API
-# should be created here. As the API evolves, models may become more
-# specific to a particular version of the API. In this case, the models
-# in question should be moved to versioned sub-package.
+from model_utils.models import TimeStampedModel
```
ashultz0, line 1: it feels weird to have these in the API directory

I would expect items in the API vs the core to be specifically tailored to API use - for example combining attributes from three DB models into a single domain concept for the front end, or providing a simplified view of a database model, or whatever

ashultz0: or maybe because core has the user model which connects back to the rest of edx these models deserve their own exam app. or the user model should be evicted into a separate user app

edx_exams/apps/core/models.py
alangsto: Makes sense, will remove these fields

edx_exams/apps/api/models.py
alangsto: Looking at other apps, the models are just all grouped within core/models.py. I'll do the same here.

MichaelRoytman (approved):
The PyLint error fixes look good to me. Thanks for fixing those!

ashultz0 (approved):

Zacharis278 (approved):
LGTM!

PR #16, opened 2022-06-14
feat: add patch endpoint to update exams

alangsto:
**JIRA:** https://2u-internal.atlassian.net/browse/MST-1518

**Description:** Added a patch endpoint to modify/update a list of exams.

The endpoint expects a list of exam data. Given this list of exam data for a course, this view will either create a new exam (if one doesn't exist), or modify an existing exam. Any course exams missing from the request will be marked as inactive.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**


edx_exams/apps/api/serializers.py
```diff
@@ -1,4 +1,33 @@
...
+    """
+    Serializer for the Exam Model
+    """
+
+    exam_name = serializers.CharField(required=True)
+    course_id = serializers.CharField(required=True)
+    content_id = serializers.CharField(required=True)
+    time_limit_mins = serializers.IntegerField(required=True)
+    due_date = serializers.DateTimeField(required=False, format=None)
+    exam_type = serializers.CharField(required=True)
```
alangsto, line 19: Do we need extra validation on exam_type? Right now it could be any string. I feel like it should be limited to a set, but I'm not sure what that set should contain.

edx_exams/apps/api/test_utils/mixins.py
```diff
@@ -0,0 +1,51 @@
...
+
+        return {
+            "iss": self.JWT_ISSUER,
+            "sub": user.pk,
+            "aud": self.JWT_AUDIENCE,
+            "nonce": "dummy-nonce",
+            "exp": now + ttl,
+            "iat": now,
+            "preferred_username": user.username,
+            "administrator": user.is_staff,
```
alangsto, line 45: Right now my tests are only working if admin is True, which shouldn't be the case. Leaving this in for now so the tests can run, but I need to figure out what's going on with JWT.

edx_exams/apps/api/serializers.py
```diff
@@ -1,4 +1,33 @@
...
+from edx_exams.apps.core.models import Exam
+
+
+class ExamSerializer(serializers.ModelSerializer):
+    """
+    Serializer for the Exam Model
+    """
+
+    exam_name = serializers.CharField(required=True)
+    course_id = serializers.CharField(required=True)
```
alangsto, line 15: There's no validation that all of the course_id's within a request have to match. Should we validate that?

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,164 @@
...
+                visited_exams.append(exam['content_id'])
+
+                if exam_object:
+                    if exam['exam_type'] != exam_object.exam_type:
+                        # if the exam type has changed, mark existing exam as inactive and then create
+                        # a new exam
+                        exam_object.is_active = False
+                        exam_object.save()
+
+                        current = Exam.objects.create(
```
alangsto, line 87: I could move some of this logic outside of the view to make this easier to read

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,164 @@
...
+                    # check that configuration exists
+                    config = CourseExamConfiguration.objects.filter(course_id=exam['course_id'])
+
+                    if len(config) > 0:
+                        provider = config[0].provider
+
+                    else:
+                        # if configuration doesn't already request, return 400
+                        error_msg = "Cannot update exam. No exam configuration set up for course"
+                        return Response(error_msg, status=status.HTTP_400_BAD_REQUEST)
```
alangsto, line 133: Decided to return 400 if there is no config. If we want, we could create a default config, but not sure what those values would be, i.e. who is our default provider?

edx_exams/apps/api/test_utils/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Base class for API Tests
+    """
+
+    def setUp(self):
+        """
+        Perform operations common to all tests.
+        """
+        super().setUp()
+        self.create_user(username=TEST_USERNAME, email=TEST_EMAIL, password=TEST_PASSWORD, is_staff=True)
```
alangsto, line 22: is_staff is being changed to false somewhere, which is why I'm getting a 403 response.

edx_exams/apps/api/serializers.py
schenedx: That's not something you can do in this service right? Leave that to the LMS service instead.

edx_exams/apps/api/serializers.py
schenedx: That can be done later. Maybe create a jira ticket and evaluate the need for that validation later?

edx_exams/apps/api/test_utils/__init__.py
schenedx: I think the problem is that, when you are converting the user information into JWT, you did not specify the user is staff or admin in the JWT token construction?

edx_exams/apps/api/test_utils/mixins.py
```diff
@@ -0,0 +1,51 @@
...
+
+        return {
+            "iss": self.JWT_ISSUER,
+            "sub": user.pk,
+            "aud": self.JWT_AUDIENCE,
+            "nonce": "dummy-nonce",
+            "exp": now + ttl,
+            "iat": now,
+            "preferred_username": user.username,
+            "administrator": is_admin,
```
schenedx, line 45: There is no case where is_admin is set to `True`

edx_exams/apps/api/v1/views.py
schenedx: Is this a pattern you see repeated? If not, it's OK to see this here.

schenedx (commented):
I got some comments inline.

edx_exams/apps/api/v1/views.py
ashultz0: It would be easier to test this logic if the serializer were not in the loop so that there is a function just over the list of exams and I do think it's a lot of stuff that doesn't need to be in the view.

This could be:
view extracts exams and calls update_these_here_exams(exams)
update separates the list to call update_exam(exam) or create_exam(exam)

most of the interesting tests are in update_exam to handle the update and maybe retire old one

edx_exams/apps/api/test_utils/__init__.py
alangsto: Got it, so administrator should be `True`. Will update that.

edx_exams/apps/api/serializers.py
ashultz0: the URL already gives us the course ID, there isn't actually a need for it in the serialized object

alangsto: I like that, I'll remove it here. I wanted to keep it in the URL in case someone sends an empty request (which should then mark all exams as inactive).

edx_exams/apps/api/v1/views.py
ashultz0: 😆 I meant the serializer was not "in the loop" in the sense of being involved in the most interesting function to test. What an unfortunate choice of phrase.

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,171 @@
...
+
+        serializer = ExamSerializer(data=request_exams, many=True)
+
+        if serializer.is_valid():
+            course_exams = Exam.objects.filter(course_id=course_id)
+
+            visited_exams = []
+
+            for exam in request_exams:
+                exam_object = course_exams.filter(content_id=exam['content_id'])
```
ashultz0, line 82: maybe "existing_exam"

edx_exams/apps/api/v1/views.py
Zacharis278: There's definitely a user flow thing here where we have to decide if we want course teams to be able to publish timed/proctored exams prior to explicitly choosing a provider. Which gets me thinking what about non-proctored exams these should probably be set to null anyway 🤔

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,171 @@
...
+                    log.info(
+                        "Created new exam=%(current_exam_id)s",
+                        {
+                            'current_exam_id': current.id,
+                        }
+                    )
+
+            # mark any exams not included in the request as inactive. The Query set has already been filtered by course
+            remaining_exams = course_exams.exclude(content_id__in=visited_exams)
+            remaining_exams.update(is_active=False)
```
ashultz0, line 163: I think this might mark any new exams and is_active = false
changed exams are ok even though the exam object is new because they have the same content ID

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,171 @@
...
+    def patch(self, request, course_id):
+        """ create or update an exam """
+        request_exams = request.data
+
+        serializer = ExamSerializer(data=request_exams, many=True)
+
+        if serializer.is_valid():
+            course_exams = Exam.objects.filter(course_id=course_id)
+
+            visited_exams = []
```
ashultz0, line 79: maybe active_exams or still_relevant_exams or dont_mark_inactive

actually it's not even exams in here
content_with_exam maybe

edx_exams/apps/api/v1/views.py
alangsto: It shouldn't, because any content ID (whether it's new or existing) is added to visited_exams

ashultz0: yeah I was just reading again and realized visited is in the way outer serializer loop so my original comment was just wrong. Although maybe fuel to your feeling that there is a bit too much in this function.

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,180 @@
...
+from edx_exams.apps.core.exam_types import get_exam_type
+from edx_exams.apps.core.models import CourseExamConfiguration, Exam
+
+
+log = logging.getLogger(__name__)
+
+
+class ExamsView(APIView):
+    """
+    View to modify exams.
```
Zacharis278, line 24: I think the comments could be a bit clearer calling out this endpoint modifies all exams for a course.

edx_exams/apps/api/v1/views.py
alangsto: The provider is now set to None if the exam is either timed, or if it is set to proctored but there is no config set up for the course. 

Thoughts on allowing None for the second case? Do we want a placeholder config (what would that look like)?

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,184 @@
...
+
+from edx_exams.apps.api.permissions import StaffUserPermissions
+from edx_exams.apps.api.serializers import ExamSerializer
+from edx_exams.apps.core.exam_types import get_exam_type
+from edx_exams.apps.core.models import CourseExamConfiguration, Exam
+
+log = logging.getLogger(__name__)
+
+
+class ExamsView(APIView):
```
ashultz0, line 21: Zach's last comment made me realize this is actually the CourseExamsView - you can't use it to operate on exams outside a course or even a single exam in a course

edx_exams/apps/api/v1/views.py
Zacharis278: None sounds fine for the moment, we'll probably have to revisit this later.

edx_exams/apps/core/models.py
```diff
@@ -97,6 +104,7 @@ class Meta:
         """ Meta class for this Django model """
         db_table = 'exams_exam'
         verbose_name = 'exam'
+        unique_together = (('course_id', 'content_id', 'exam_type', 'provider'),)
```
alangsto, line 107: Explicitly making this combination unique

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,258 @@
...
+import pytz
+from django.urls import reverse
+
+from edx_exams.apps.api.test_utils import ExamsAPITestCase
+from edx_exams.apps.api.test_utils.factories import UserFactory
+from edx_exams.apps.core.models import CourseExamConfiguration, Exam
+
+
+@ddt.ddt
+class ExamViewTests(ExamsAPITestCase):
```
ashultz0, line 18: because I was pedantic about the view name this now also needs to be renamed

there is no reward for virtue

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,175 @@
...
+
+    @classmethod
+    def handle_exams(cls, request_exams_list, course_exams_qs, course_id):
+        """
+        Decide how exams should be updated or created
+        """
+
+        for exam in request_exams_list:
+            # should only be one object per exam type per content_id
+            type_exam_qs = course_exams_qs.filter(content_id=exam['content_id'], exam_type=exam['exam_type'])
```
ashultz0, line 92: is this going to do one SQL query per exam in the request list? maybe it would make sense to get all the course exams into a dict by content ID at the top of the loop and use that

ashultz0 (changes requested):
this is a lot cleaner and easier to follow now

I think there is one N+1 query we can avoid (unless I'm wrong and django would be smart here but I don't see how it could be)

edx_exams/apps/api/v1/views.py
alangsto: Created dict to organize info 👍

edx_exams/apps/api/v1/views.py
```diff
@@ -1 +1,175 @@
...
+
+            existing_type_exam = type_exam_qs[0] if len(type_exam_qs) > 0 else None
```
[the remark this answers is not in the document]
alangsto: Replaced this with exam dictionary, so filtering isn't needed anymore. I did, however, have similar logic further down so I updated it there

alangsto:
@ashultz0 @Zacharis278 thank you for the very helpful feedback! I incorporated the suggested changes into my latest commit

ashultz0 (approved):
👍 fixes my concerns

PR #19, opened 2022-06-23
feat: carry fields from LMS user into our local user

ashultz0:
Including saving the universal LMS ID identifier locally. Without this setting only email carries through.

This setup and these settings are modeled on what the demographics service is using. We will need to follow this with updates to our edx.org settings.


alangsto (approved):
👍  one question but this looks good

PR #24, opened 2022-07-25
add course exam config endpoint

varshamenon4:
**JIRA:** [WFDASH-1523](https://2u-internal.atlassian.net/browse/MST-1523)

**Description:** Creates an api endpoint that creates/updates course exam configs given the course ID and the proctoring provider.

migrations, etc.

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**
- [ ] All reviewers approved
- [x] CI build is green
- [ ] Changelog record added
- [x] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -11,7 +11,7 @@ from django.urls import reverse
 
 from edx_exams.apps.api.test_utils import ExamsAPITestCase
 from edx_exams.apps.api.test_utils.factories import UserFactory
-from edx_exams.apps.core.models import CourseExamConfiguration, Exam
+from edx_exams.apps.core.models import CourseExamConfiguration, Exam, ProctoringProvider
 
 
 @ddt.ddt
@@ -256,3 +256,93 @@ class CourseExamsViewTests(ExamsAPITestCase):
         self.assertEqual(new_exam.due_date, pytz.utc.localize(datetime.fromisoformat('2025-07-01 00:00:00')))
         self.assertEqual(new_exam.hide_after_due, False)
         self.assertEqual(new_exam.is_active, True)
+
+
+@ddt.ddt
+class CourseExamConfigurationsViewTests(ExamsAPITestCase):
+    """
+    Test CourseExamConfigurationsView
+    """
+
+    def setUp(self):
+        super().setUp()
+
+        self.course_id = 'course-v1:edx+test+f19'
+
+        self.url = reverse('api:v1:course-exam-config', kwargs={'course_id': self.course_id})
+
+    def patch_api(self, user, data):
+        """
+        Helper function to make patch request to the API
+        """
+
+        data = json.dumps(data)
+        headers = self.build_jwt_headers(user)
+
+        return self.client.patch(self.url, data, **headers, content_type="application/json")
+
+    def test_auth_failures(self):
+        """
+        Verify the endpoint validates permissions
+        """
+        # Test unauthenticated
+        response = self.client.patch(self.url)
+        self.assertEqual(response.status_code, 401)
+
+        # Test non-staff user
+        random_user = UserFactory()
+        response = self.patch_api(random_user, {})
+        self.assertEqual(403, response.status_code)
+
+    def test_invalid_data(self):
+        """
+        Assert that endpoint returns 400 if provider is missing
+        """
+        data = {}
+
+        response = self.patch_api(self.user, data)
+        self.assertEqual(400, response.status_code)
+
+    def test_invalid_provider(self):
+        """
+        Assert endpoint returns 400 if provider is invalid
+        """
+        data = {'provider': 'nonexistent_provider'}
+
+        response = self.patch_api(self.user, data)
+        self.assertEqual(400, response.status_code)
+
+    def test_config_update(self):
+        """
+        Test that config is updated
+        """
+        CourseExamConfiguration.objects.create(
+            course_id=self.course_id,
+            provider=self.test_provider
+        )
+        provider = ProctoringProvider.objects.create(
+            name='test_provider_2',
+            verbose_name='testing_provider_2',
+            lti_configuration_id='223456789'
+        )
+        data = {'provider': provider.name}
+
+        response = self.patch_api(self.user, data)
+        self.assertEqual(204, response.status_code)
+        self.assertEqual(len(CourseExamConfiguration.objects.all()), 1)
+
+        config = CourseExamConfiguration.objects.get(course_id=self.course_id)
+        self.assertEqual(config.provider, provider)
+
+    def test_config_create(self):
+        """
+        Test that config is created
+        """
+        data = {'provider': 'test_provider'}
+
+        response = self.patch_api(self.user, data)
+        self.assertEqual(204, response.status_code)
+        self.assertEqual(len(CourseExamConfiguration.objects.all()), 1)
+
+        config = CourseExamConfiguration.objects.get(course_id=self.course_id)
+        self.assertEqual(config.provider, self.test_provider)
--- a/edx_exams/apps/api/v1/urls.py
+++ b/edx_exams/apps/api/v1/urls.py
@@ -2,11 +2,16 @@
 
 from django.urls import re_path
 
-from edx_exams.apps.api.v1.views import CourseExamsView
+from edx_exams.apps.api.v1.views import CourseExamConfigurationsView, CourseExamsView
 from edx_exams.apps.core.constants import COURSE_ID_PATTERN
 
 app_name = 'v1'
 
 urlpatterns = [
-    re_path(fr'exams/course_id/{COURSE_ID_PATTERN}', CourseExamsView.as_view(), name='exams-course_exams'),
+    re_path(fr'exams/course_id/{COURSE_ID_PATTERN}',
+            CourseExamsView.as_view(),
+            name='exams-course_exams'),
+    re_path(fr'configs/course_id/{COURSE_ID_PATTERN}',
+            CourseExamConfigurationsView.as_view(),
+            name='course-exam-config')
 ]
--- a/edx_exams/apps/api/v1/views.py
+++ b/edx_exams/apps/api/v1/views.py
@@ -4,6 +4,7 @@ V1 API Views
 import logging
 import uuid
 
+from django.core.exceptions import ObjectDoesNotExist
 from edx_api_doc_tools import path_parameter, schema
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from rest_framework import status
@@ -13,7 +14,7 @@ from rest_framework.views import APIView
 from edx_exams.apps.api.permissions import StaffUserPermissions
 from edx_exams.apps.api.serializers import ExamSerializer
 from edx_exams.apps.core.exam_types import get_exam_type
-from edx_exams.apps.core.models import CourseExamConfiguration, Exam
+from edx_exams.apps.core.models import CourseExamConfiguration, Exam, ProctoringProvider
 
 log = logging.getLogger(__name__)
 
@@ -177,5 +178,58 @@ class CourseExamsView(APIView):
         else:
             response_status = status.HTTP_400_BAD_REQUEST
             data = {"detail": "Invalid data", "errors": serializer.errors}
+
+        return Response(status=response_status, data=data)
+
+
+class CourseExamConfigurationsView(APIView):
+    """
+    View to create and update course exam configs for a specific course.
+
+    Given a course id and a proctoring provider name, this view will either create a new course exam configuration
+    (if one doesn't exist), or modify the proctoring provider on an existing course exam config.
+
+    HTTP PATCH
+    Creates or updates a CourseExamConfiguration.
+    Expected PATCH data: {
+        'provider': 'test_provider',
+    }
+    **PATCH data Parameters**
+        * name: This is the name of the proctoring provider.
+    **Exceptions**
+        * HTTP_400_BAD_REQUEST
+    """
+
+    authentication_classes = (JwtAuthentication,)
+    permission_classes = (StaffUserPermissions,)
+
+    @classmethod
+    def handle_config(cls, provider, course_id):
+        """
+        Helper method that decides whether to update existing or create new config.
+        """
+        CourseExamConfiguration.objects.update_or_create(
+            course_id=course_id,
+            defaults={'provider': provider})
+        log.info(f"Created or updated course exam configuration course_id={course_id},provider={provider.name}")
+
+    def patch(self, request, course_id):
+        """
+        Create/update course exam configuration.
+        """
+        # check that proctoring provider is in request
+        if request.data.get('provider') is None:
+            response_status = status.HTTP_400_BAD_REQUEST
+            data = {"detail": "No proctoring provider name in request."}
+        else:
+            try:
+                provider = ProctoringProvider.objects.get(name=request.data['provider'])
+                self.handle_config(provider, course_id)
+                response_status = status.HTTP_204_NO_CONTENT
+                data = {}
+            # return 400 if proctoring provider does not exist
+            except ObjectDoesNotExist:
+                response_status = status.HTTP_400_BAD_REQUEST
+                data = {"detail": "Proctoring provider does not exist."}
 
         return Response(status=response_status, data=data)
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -72,7 +72,7 @@ django-model-utils==4.2.0
     # via -r requirements/base.in
 django-rest-swagger==2.2.0
     # via -r requirements/base.in
-django-waffle==2.5.0
+django-waffle==2.6.0
     # via
     #   -r requirements/base.in
     #   edx-django-utils
@@ -88,7 +88,7 @@ djangorestframework==3.13.1
     #   edx-drf-extensions
 drf-jwt==1.19.2
     # via edx-drf-extensions
-drf-yasg==1.20.0
+drf-yasg==1.21.3
     # via edx-api-doc-tools
 edx-api-doc-tools==1.6.0
     # via -r requirements/base.in
@@ -143,7 +143,7 @@ markupsafe==2.1.1
     #   xblock
 mysqlclient==2.1.1
     # via -r requirements/base.in
-newrelic==7.14.0.177
+newrelic==7.16.0.178
     # via edx-django-utils
 oauthlib==3.2.0
     # via
@@ -194,6 +194,7 @@ pytz==2022.1
     #   -r requirements/base.in
     #   django
     #   djangorestframework
+    #   drf-yasg
     #   xblock
 pyyaml==6.0
     # via
@@ -239,7 +240,7 @@ social-auth-core==4.3.0
     #   social-auth-app-django
 sqlparse==0.4.2
     # via django
-stevedore==3.5.0
+stevedore==4.0.0
     # via
     #   edx-django-utils
     #   edx-opaque-keys
@@ -247,7 +248,7 @@ uritemplate==4.1.1
     # via
     #   coreapi
     #   drf-yasg
-urllib3==1.26.10
+urllib3==1.26.11
     # via requests
 web-fragments==2.0.0
     # via
--- a/requirements/ci.txt
+++ b/requirements/ci.txt
@@ -10,9 +10,9 @@ charset-normalizer==2.1.0
     # via requests
 codecov==2.1.12
     # via -r requirements/ci.in
-coverage==6.4.1
+coverage==6.4.2
     # via codecov
-distlib==0.3.4
+distlib==0.3.5
     # via virtualenv
 filelock==3.7.1
     # via
@@ -40,7 +40,7 @@ toml==0.10.2
     # via tox
 tox==3.25.1
     # via -r requirements/ci.in
-urllib3==1.26.10
+urllib3==1.26.11
     # via requests
 virtualenv==20.15.1
     # via tox
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -77,7 +77,7 @@ coreschema==0.0.4
     #   -r requirements/validation.txt
     #   coreapi
     #   drf-yasg
-coverage[toml]==6.4.1
+coverage[toml]==6.4.2
     # via
     #   -r requirements/validation.txt
     #   pytest-cov
@@ -85,7 +85,6 @@ cryptography==37.0.4
     # via
     #   -r requirements/validation.txt
     #   pyjwt
-    #   secretstorage
     #   social-auth-core
 ddt==1.5.0
     # via -r requirements/validation.txt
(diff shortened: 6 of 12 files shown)


openedx-webhooks:
Thanks for the pull request, @varshamenon4!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.




:warning: We can't start reviewing your pull request until you've submitted a     [signed contributor agreement](https://openedx.org/cla)     or indicated your institutional affiliation. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.     If you've signed an agreement in the past, you may need to re-sign.  See     [The New Home of the Open edX Codebase](https://open.edx.org/blog/the-new-home-of-the-open-edx-codebase/)     for details.

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,78 @@ def patch(self, request, course_id):
...
+
+    authentication_classes = (JwtAuthentication,)
+    permission_classes = (StaffUserPermissions,)
+
+    @classmethod
+    def handle_config(cls, provider, course_id):
+        """
+        Helper method that decides whether to update existing or create new config.
+        """
+        # todo: is there a better way to check whether the config already exists?
```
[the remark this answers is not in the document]
alangsto: You could use one of the `_or_create` django methods, that way you can either update the existing object or just create a new one.

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,78 @@ def patch(self, request, course_id):
...
+        # todo: does the course_id need a uniqueness constraint in the CourseExamConfiguration model?
```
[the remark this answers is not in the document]
alangsto: I think there should only be one config per course

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,78 @@ def patch(self, request, course_id):
...
+    def patch(self, request, course_id):
+        """
+        Create/update course exam configuration.
+        """
+
+        serializer = ProctoringProviderSerializer(data=request.data)
+
+        if serializer.is_valid():
+            provider = serializer.save()
+            self.handle_config(provider, course_id)
```
alangsto, line 248: Part of this ticket is to determine if we also need to update a course's exams after the config is changed. I do think we need that behavior, so it should be handled as part of this view.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,101 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+        data = {
+            'name': 'test_provider2',
+            'verbose_name': 'testing_provider_2',
+            'lti_configuration_id': '223456789'
+        }
+
+        self.assertEqual(len(CourseExamConfiguration.objects.all()), 1)
+        self.get_response(self.user, data, 200)
+        self.assertEqual(len(CourseExamConfiguration.objects.all()), 1)
+        # todo: should I test that the config is correct? anything else to test?
```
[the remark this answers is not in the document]
alangsto: Yes, if you are updating the content, you should check that the expected updates are carried through

edx_exams/apps/api/v1/views.py
varshamenon4: Yes, you're right @alangsto! So @Zacharis278 had mentioned maybe needing to wait on this until [this decision was made](https://2u-internal.atlassian.net/wiki/spaces/PT/pages/95223955/Discovery+Altering+Active+Exams). In the meantime, do you think it's best to move forward as is for this ticket then create a new ticket to add this behavior?

edx_exams/apps/api/v1/urls.py
```diff
@@ -2,11 +2,16 @@
...
 from edx_exams.apps.core.constants import COURSE_ID_PATTERN
 
 app_name = 'v1'
 
 urlpatterns = [
-    re_path(fr'exams/course_id/{COURSE_ID_PATTERN}', CourseExamsView.as_view(), name='exams-course_exams'),
+    re_path(fr'exams/course_id/{COURSE_ID_PATTERN}',
+            CourseExamsView.as_view(),
+            name='exams-course_exams'),
+    re_path(fr'configs/course_id/{COURSE_ID_PATTERN}',
```
Zacharis278, line 14: NIT, but 'configs' to me reads as plural. I'd just go with configuration or config

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,78 @@ def patch(self, request, course_id):
...
+
+    def patch(self, request, course_id):
+        """
+        Create/update course exam configuration.
+        """
+
+        serializer = ProctoringProviderSerializer(data=request.data)
+
+        if serializer.is_valid():
+            provider = serializer.save()
```
Zacharis278, line 247: So we shouldn't need to actually save/modify any providers is part of this endpoint. This should simply create a configuration object that references one.

The use case for this endpoint is it's going to be called when a new value is selected here: https://course-authoring.stage.edx.org/course/course-v1:edX+cheating101+2018T3/pages-and-resources/proctoring/settings

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,78 @@ def patch(self, request, course_id):
...
+    def handle_config(cls, provider, course_id):
+        """
+        Helper method that decides whether to update existing or create new config.
+        """
+        # todo: is there a better way to check whether the config already exists?
+        # todo: does the course_id need a uniqueness constraint in the CourseExamConfiguration model?
+        config = CourseExamConfiguration.objects.filter(course_id=course_id)
+
+        # update if config exists
+        if len(config) == 1:
```
MichaelRoytman, line 220: [exists](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#exists) is supposedly a little faster than checking for a non-empty QuerySet this way. But this is moot if you go with Alie's advice about the `update_or_create` method.

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,78 @@ def patch(self, request, course_id):
...
+            config[0].provider = provider
+            config[0].save()
+            log.info("Updated course exam configuration provider=%(provider_name)s",
```
MichaelRoytman, line 223: Nit. 

If you want, I believe you can also do...
```suggestion
            log.info(f"Updated course exam configuration provider={provider.name}",
```

or 

```suggestion
            provider_name = provider.name
            log.info(f"Updated course exam configuration provider={provider_name}",
```

I'm just suggesting this because you can probably fit that all on one line.

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,78 @@ def patch(self, request, course_id):
...
+                         'provider_name': provider.name,
+                     }
+                     )
+
+    def patch(self, request, course_id):
+        """
+        Create/update course exam configuration.
+        """
+
+        serializer = ProctoringProviderSerializer(data=request.data)
```
MichaelRoytman, line 244: Based on Zach's feedback, you won't need a serializer, since you're not saving the `ProctoringProvider` here. But it would make sense to handle the case that the `request.data['provider']` does not correspond to an existing `ProctoringProvider`.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,101 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+        Test that config is created
+        """
+        data = {
+            'name': 'test_provider',
+            'verbose_name': 'testing_provider',
+            'lti_configuration_id': '123456789'
+        }
+        self.assertEqual(len(CourseExamConfiguration.objects.all()), 0)
+        self.get_response(self.user, data, 200)
+        self.assertEqual(len(CourseExamConfiguration.objects.all()), 1)
```
MichaelRoytman, line 355: I think you should test that your desired data is actually what was saved. Otherwise, this test can only tell you that _something_ was created. The only way to ensure that the `CourseExamConfiguration` that you expected was created was to test the values. For example, you could test that the `provider_id` field maps to the correct `ProctoringProvider`.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,101 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+
+        return self.client.patch(self.url, data, **headers, content_type="application/json")
+
+    # todo: can this helper method be moved to base since both views use it?
+    def get_response(self, user, data, expected_response):
+        """
+        Helper functions to get API response
+        """
+        response = self.patch_api(user, data)
+        self.assertEqual(response.status_code, expected_response)
```
MichaelRoytman, line 291: Nit: This is feedback I gave on Marcin's pull request in `program-intent-engagement`, since it borrowed some functions from `edx-exams`.

Having an assertion in a function called `get_response` is unexpected to me. I think it would be more clear what a test is actually testing if the assertion is defined within the test. `self.get_response(self.user, 200)` doesn't really tell me the purpose of the `200`. It's not clear what the purpose of the `200` is in a method called `get_response`. `get_response`'s obvious responsibility is to make a request and return the response. If it also starts asserting things about the response, it's doing more than the name claims that it does, and it becomes harder to read, in my opinion.

An alternative would be to remove the line `self.assertEqual(response.status_code, expected_response)` and just make that assertion directly in your test. That way, it's clear in the actual test what you are testing. I think it would be clearer to collocate all assertions your test needs to make within the test.

Let me know what you think!

edx_exams/apps/api/v1/views.py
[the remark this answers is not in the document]
varshamenon4: Merged this: https://github.com/edx/edx-exams/pull/28

edx_exams/apps/api/v1/tests/test_views.py
varshamenon4: Ah, this totally makes sense @MichaelRoytman ! I've updated to have the asserts be in the individual tests. Since the `get_response` method is now a little redundant, I've removed and just kept the `patch_api` helper method. Let me know if you think it should be renamed to be more clear though!

edx_exams/apps/api/v1/urls.py
varshamenon4: Updated to config!

edx_exams/apps/api/v1/urls.py
```diff
@@ -2,11 +2,16 @@
...
+            CourseExamConfigurationsView.as_view(),
+            name='exams-course_exam_configs')
```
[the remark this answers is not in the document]
varshamenon4: It was intended, but I was using `exams-course_exams` on line 13 as a reference, so I'm not sure if that's right? @MichaelRoytman

edx_exams/apps/api/v1/views.py
varshamenon4: Oh, yeah! Nice call, @MichaelRoytman! Made those changes to both this log and the one for the create log. ty!

edx_exams/apps/api/v1/views.py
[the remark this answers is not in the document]
varshamenon4: Oooh, nice @alangsto! I like this! ty. Refactored. :)

edx_exams/apps/api/v1/views.py
varshamenon4: This is really good to know, @MichaelRoytman! I ended up changing to `update_or_create` method but I'll plan on using exists moving forward if needed. :)

edx_exams/apps/api/v1/views.py
varshamenon4: Cool. So would the endpoint take in the provider name the same way as course_id? @Zacharis278

edx_exams/apps/api/v1/urls.py
varshamenon4: @Zacharis278 I was just looking at this and it seems to indicate that it should be plural: https://openedx.atlassian.net/wiki/spaces/AC/pages/18350757/edX+REST+API+Conventions
But maybe I'm misunderstanding?

MichaelRoytman: I'm reading it the same way. Plural it is, then! Thanks for double checking that.

Zacharis278: huh yeah 'configs' just struck me as odd at first glace since there is only one per course but that doc is definitely correct. Guess it should be plural. I rescind my comment!

edx_exams/apps/api/v1/views.py
Zacharis278: The provider should be part of the request body. The configuration object is likely going to have more fields in the future so this should be a dictionary of key/values. We just only have one key so far, provider.

edx_exams/apps/api/v1/tests/test_views.py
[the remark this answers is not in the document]
varshamenon4: Makes sense! Added asserts to check that expected updates are carried through.

edx_exams/apps/api/v1/urls.py
varshamenon4: Sweet! Thanks for confirming @MichaelRoytman and @Zacharis278! Will update.

edx_exams/apps/api/v1/views.py
varshamenon4: Makes sense! Thanks @MichaelRoytman. Updated logic.

edx_exams/apps/api/v1/tests/test_views.py
varshamenon4: @MichaelRoytman cool! I ended up checking that the providers are the same as expected.

edx_exams/apps/api/v1/views.py
varshamenon4: Nice, thanks for confirming! @Zacharis278

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,78 @@ def patch(self, request, course_id):
...
+    # todo: make parameters descriptions more useful.
+    """
+    View to create and update course exam configs for a specific course.
+
+    Given a course id and a proctoring provider, this view will either create a new course exam configuration
+    (if one doesn't exist), or modify the proctoring provider on an existing course exam config.
+
+    HTTP PATCH
+    Creates or updates a CourseExamConfiguration.
+    Expected PATCH data: {
```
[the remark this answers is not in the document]
varshamenon4: You're right @MichaelRoytman! I was a bit confused on that but thanks for helping clarify! So I've updated the params to only include provider name, so I'm guessing PATCH is still right?

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,93 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+            lti_configuration_id='223456789'
+        )
+        data = {'provider': provider.name}
+
+        response = self.patch_api(self.user, data)
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(len(CourseExamConfiguration.objects.all()), 1)
+
+        config = CourseExamConfiguration.objects.get(course_id=self.course_id)
+        self.assertEqual(config.provider, provider)
```
varshamenon4, line 335: I'm now checking that there is a config and that the contents are correct. Is this a good way to test that the contents of the config are correct?

edx_exams/apps/api/v1/urls.py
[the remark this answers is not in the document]
varshamenon4: I am using the above exams url as an example, but let me know if this should be different.

edx_exams/apps/api/v1/views.py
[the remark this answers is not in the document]
varshamenon4: Is patch still the right request type here?

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +180,53 @@ def patch(self, request, course_id):
...
+    def patch(self, request, course_id):
+        """
+        Create/update course exam configuration.
+        """
+        try:
+            provider = ProctoringProvider.objects.get(name=request.data['provider'])
+            self.handle_config(provider, course_id)
+            response_status = status.HTTP_200_OK
+            data = {}
+        except KeyError:
```
varshamenon4, line 225: Does the way I've handled the errors make sense here? Let me know if additional info is needed or if there is a better approach.

edx_exams/apps/api/v1/views.py
varshamenon4: I've created that ticket [here](https://2u-internal.atlassian.net/browse/MST-1579).

Let me know if anything's missing in that ticket. :)

varshamenon4 (commented):
Thanks y'all for the great feedback! Super helpful. I've made those changes and added some comments for review.

edx_exams/apps/api/v1/views.py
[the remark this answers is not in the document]
Zacharis278: I think a PUT technically makes sense because it 'replaces' the entire object. If we add additional values (such as the escalation email) I'd expect that the UI to update all of them at once.

edx_exams/apps/api/v1/urls.py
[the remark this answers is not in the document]
MichaelRoytman: I think it's fine. I think it just looks a little odd to have a mixture of _ and -, but there's nothing wrong with it. Feel free to disregard!

edx_exams/apps/api/v1/views.py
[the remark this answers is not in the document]
MichaelRoytman: I think a PUT requires that you supply a complete representation of the object to be replaced. If you're supplying just a single field (i.e. `name`), that would make it a PATCH.

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +180,53 @@ def patch(self, request, course_id):
...
+            course_id=course_id,
+            defaults={'provider': provider})
+        log.info(f"Created or updated course exam configuration course_id={course_id},provider={provider.name}")
+
+    def patch(self, request, course_id):
+        """
+        Create/update course exam configuration.
+        """
+        try:
+            provider = ProctoringProvider.objects.get(name=request.data['provider'])
```
MichaelRoytman, line 221: The `ProctoringProvider` doesn't have a uniqueness constraint on `name`, so it's possible for this to return more than one provider. We should either add a uniqueness constraint on the field or handle the case where more than one provider is returned.

edx_exams/apps/api/v1/views.py
MichaelRoytman: I think this is good. You could also just proactively check for this value at the start of the view. 

```if not request.data.get('provider'):
         response_status = status.HTTP_400_BAD_REQUEST
         data = {"detail": "No proctoring provider name in request."}
```

The `get` will return `None` if `provider` isn't in `request.data`.

The difference is small. I'd say the advantage of doing it this way is that it's easier to read. It took me a second to figure out why you were handling a `KeyError`. I had to read the response data and scroll back up to see what could throw a `KeyError`. It's not a big deal, though. I think this is good.

You could do it in a serializer, as an alternative, but that might be overkill for something like this.

edx_exams/apps/api/v1/tests/test_views.py
MichaelRoytman: I think this looks good!

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +180,53 @@ def patch(self, request, course_id):
...
+            self.handle_config(provider, course_id)
+            response_status = status.HTTP_200_OK
```
MichaelRoytman, line 223: I _think_ it would be more accurate for this to be `204`, per this [document](https://datatracker.ietf.org/doc/html/rfc5789#section-2.1), but I'm not positive.

edx_exams/apps/api/v1/views.py
varshamenon4: @MichaelRoytman, I like this approach! My thought with the `KeyError` was having both errors handled in the try except block but I agree that that approach is less readable. I've gone ahead and updated.

edx_exams/apps/api/v1/urls.py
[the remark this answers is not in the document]
varshamenon4: @MichaelRoytman, I was curious why it was like that so just looked at registrar for another example and it looks like only dashes are used. So I just changed to use only dashes for the course exam config url. For consistency, should I go ahead and change for the course exams name? Doesn't seem like it'll impact the user but just for consistency/a nit?

edx_exams/apps/api/v1/views.py
varshamenon4: That's a good point! I can create another PR as I did for adding the uniqueness constraint to the course exam config.

edx_exams/apps/api/v1/views.py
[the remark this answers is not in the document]
Zacharis278: So the way I'm looking at this CourseExamConfiguration is the resource being replaced, ProctoringProvider is just a related object.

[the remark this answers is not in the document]
varshamenon4: Based on [this document](https://openedx.atlassian.net/wiki/spaces/AC/pages/18350757/edX+REST+API+Conventions), `PATCH` seems a better fit to me as well. Does that sound good, @Zacharis278? Definitely open to thoughts on the advantages to using `PUT` here. Thanks Zach and @MichaelRoytman for the thoughts here!

edx_exams/apps/api/v1/views.py
varshamenon4: Oh, yes, that's right! Thanks for catching that @MichaelRoytman! Updated.

Zacharis278: fwiw 200 is representative of 2xx responses it's just less specific. Since the UI isn't going to care about the distinction here I don't think it matters what one we use.

edx_exams/apps/api/v1/views.py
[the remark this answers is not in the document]
Zacharis278: yeah, I suppose PATCH is a better catch-all anyway

edx_exams/apps/api/v1/views.py
varshamenon4: Here's [that PR](https://github.com/edx/edx-exams/pull/30)!

edx_exams/apps/api/v1/views.py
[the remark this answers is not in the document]
varshamenon4: Sounds good!

edx_exams/apps/api/v1/views.py
varshamenon4: Merged the above!

Zacharis278 (approved):
🚀

alangsto (approved):
👍

PR #27, opened 2022-07-25
feat: adding get endpoint for proctoring providers

m-pitera:
**JIRA:** [MST-1522](https://2u-internal.atlassian.net/browse/MST-1522)

**Description:** This adds a get endpoint to retrieve a list of all available proctoring providers

### DIFF
--- a/edx_exams/apps/api/serializers.py
+++ b/edx_exams/apps/api/serializers.py
@@ -4,7 +4,7 @@ Serializers for the edx-exams API
 from rest_framework import serializers
 
 from edx_exams.apps.core.exam_types import EXAM_TYPES
-from edx_exams.apps.core.models import Exam
+from edx_exams.apps.core.models import Exam, ProctoringProvider
 
 
 class ExamSerializer(serializers.ModelSerializer):
@@ -41,3 +41,13 @@ class ExamSerializer(serializers.ModelSerializer):
         if value not in valid_exam_types:
             raise serializers.ValidationError("Must be a valid exam type.")
         return value
+
+
+class ProctoringProviderSerializer(serializers.ModelSerializer):
+    """
+        Serializer for the ProctoringProvider Model
+    """
+
+    class Meta:
+        model = ProctoringProvider
+        fields = "__all__"
--- a/edx_exams/apps/api/serializers.py
+++ b/edx_exams/apps/api/serializers.py
@@ -50,4 +50,4 @@ class ProctoringProviderSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = ProctoringProvider
-        fields = "__all__"
+        fields = ["name", "verbose_name", "lti_configuration_id"]
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -368,15 +368,11 @@ class ProctoringProvidersViewTest(ExamsAPITestCase):
             lti_configuration_id='223456789'
         )
 
-        # import pudb; pu.db
-
         response = self.get_response()
-        proctoring_providers_list = ProctoringProvider.objects.all()
 
         expected_data = [self.test_provider.verbose_name, test_provider2.verbose_name]
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(proctoring_providers_list), 2)
         self.assertIn(response.data[0]['verbose_name'], expected_data)
         self.assertIn(response.data[1]['verbose_name'], expected_data)
 
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -375,11 +375,9 @@ class ProctoringProvidersViewTest(ExamsAPITestCase):
 
         response_data_list.sort()
 
-        expected_data = sorted([self.test_provider.name, test_provider2.name])
-
         self.assertEqual(response.status_code, 200)
-        self.assertIn(response_data_list[0], expected_data[0])
-        self.assertIn(response_data_list[1], expected_data[1])
+        self.assertIn(response_data_list[0], self.test_provider.name)
+        self.assertIn(response_data_list[1], test_provider2.name)
 
     def test_proctoring_providers_list_empty(self):
 
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -363,9 +363,16 @@ class ProctoringProvidersViewTest(ExamsAPITestCase):
 
     def test_proctoring_providers_list(self):
 
+        test_provider2 = ProctoringProvider.objects.create(
+            name='test_provider2',
+            verbose_name='testing_provider2',
+            lti_configuration_id='223456789'
+        )
+
         response = self.get_response()
         proctoring_providers_list = ProctoringProvider.objects.all()
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(proctoring_providers_list), 1)
-        self.assertEqual(proctoring_providers_list[0], self.test_provider)
+        self.assertEqual(len(proctoring_providers_list), 2)
+        self.assertIn(self.test_provider, proctoring_providers_list)
+        self.assertIn(test_provider2, proctoring_providers_list)
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -367,14 +367,19 @@ class ProctoringProvidersViewTest(ExamsAPITestCase):
             verbose_name='testing_provider2',
             lti_configuration_id='223456789'
         )
-
         response = self.get_response()
+        response_data_list = []
 
-        expected_data = [self.test_provider.verbose_name, test_provider2.verbose_name]
+        for item in response.data:
+            response_data_list.append(item['name'])
+
+        response_data_list.sort()
+
+        expected_data = sorted([self.test_provider.name, test_provider2.name])
 
         self.assertEqual(response.status_code, 200)
-        self.assertIn(response.data[0]['verbose_name'], expected_data)
-        self.assertIn(response.data[1]['verbose_name'], expected_data)
+        self.assertIn(response_data_list[0], expected_data[0])
+        self.assertIn(response_data_list[1], expected_data[1])
 
     def test_proctoring_providers_list_empty(self):
 
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -368,20 +368,23 @@ class ProctoringProvidersViewTest(ExamsAPITestCase):
             lti_configuration_id='223456789'
         )
 
+        # import pudb; pu.db
+
         response = self.get_response()
         proctoring_providers_list = ProctoringProvider.objects.all()
 
+        expected_data = [self.test_provider.verbose_name, test_provider2.verbose_name]
+
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(proctoring_providers_list), 2)
-        self.assertIn(self.test_provider, proctoring_providers_list)
-        self.assertIn(test_provider2, proctoring_providers_list)
+        self.assertIn(response.data[0]['verbose_name'], expected_data)
+        self.assertIn(response.data[1]['verbose_name'], expected_data)
 
     def test_proctoring_providers_list_empty(self):
 
         self.test_provider.delete()
 
         response = self.get_response()
-        proctoring_providers_list = ProctoringProvider.objects.all()
 
         self.assertEqual(response.status_code, 200)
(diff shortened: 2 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @m-pitera!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.


:warning: We can't start reviewing your pull request until you've submitted a     [signed contributor agreement](https://openedx.org/cla)     or indicated your institutional affiliation. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.     If you've signed an agreement in the past, you may need to re-sign.  See     [The New Home of the Open edX Codebase](https://open.edx.org/blog/the-new-home-of-the-open-edx-codebase/)     for details.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,34 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
         self.assertEqual(new_exam.due_date, pytz.utc.localize(datetime.fromisoformat('2025-07-01 00:00:00')))
         self.assertEqual(new_exam.hide_after_due, False)
         self.assertEqual(new_exam.is_active, True)
+
+
+@ddt.ddt
```
ashultz0, line 261: I don't think we need the data driven testing decorator on the class if the tests don't actually use any ddt decorators

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,34 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+
+    def get_response(self):
+        """
+        Helper function to make a get request
+        """
+        url = reverse("api:v1:proctoring-providers-list")
+        response = self.client.get(url)
+        return response
+
+    def test_proctoring_providers_list(self):
```
ashultz0, line 275: formatting/readability nit - this def has one newline above and one below which makes it hard to see what it goes with

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,34 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+            lti_configuration_id='223456789'
+        )
+
+        response = self.get_response()
+        proctoring_providers_list = ProctoringProvider.objects.all()
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(proctoring_providers_list), 2)
+        self.assertIn(self.test_provider, proctoring_providers_list)
+        self.assertIn(test_provider2, proctoring_providers_list)
```
ashultz0, line 289: these providers list tests aren't enough - they are testing that the django object layer works, which we know and don't need to test

if you want to validate that the providers come out in the response, that will require reparsing the response somewhat

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,30 @@ def patch(self, request, course_id):
             data = {"detail": "Invalid data", "errors": serializer.errors}
 
         return Response(status=response_status, data=data)
+
+
+class ProctoringProvidersView(APIView):
```
MichaelRoytman, line 184: You could simplify this by inheriting from DRF's [ListAPIView](https://www.django-rest-framework.org/api-guide/generic-views/#listapiview).

edx_exams/apps/api/v1/urls.py
```diff
@@ -2,11 +2,20 @@
...
 app_name = 'v1'
 
 urlpatterns = [
-    re_path(fr'exams/course_id/{COURSE_ID_PATTERN}', CourseExamsView.as_view(), name='exams-course_exams'),
+    re_path(
+        fr'exams/course_id/{COURSE_ID_PATTERN}',
+        CourseExamsView.as_view(),
+        name='exams-course_exams'),
+    re_path(
+        r"^providers/available?$",
```
MichaelRoytman, line 16: I don't think you need `available` in the URL name. I think it's understood that whatever is returned by the API is what's available.

edx_exams/apps/api/v1/tests/test_views.py
m-pitera: hm ok, would this be like checking the lti_config_id of the providers?

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +179,30 @@ def patch(self, request, course_id):
...
+    def get(self, request):
+        """
+        This view should return a list of all available proctoring providers
+        """
+
+        proctoring_providers = ProctoringProvider.objects.all()
+        proctoring_providers_list = list(proctoring_providers.values())
+
+        response_status = status.HTTP_200_OK
+        data = proctoring_providers_list
```
MichaelRoytman, line 206: The line just renames the variable, and it's not really necessary. You can save yourself this step by passing it directly when creating the `Response` object, like `return Response(status=response_status, data= proctoring_providers_list)`.

edx_exams/apps/api/v1/urls.py
m-pitera: ok that makes sense

edx_exams/apps/api/v1/tests/test_views.py
ashultz0: you just need to get what you are testing from the response, not from a separately queried list

MichaelRoytman: You could hardcode exactly what you are testing. For example,

```expected_data = [{"name": "provider", "verbose_name": "verbose_provider", ...}, ...]`

You can then use assertEqual. Although, you may run into issues with the order of the list. But that's a solvable problem.

File: edx_exams/apps/api/v1/views.py (hunk 4 above)
Comment (m-pitera) [reply edx_exams/apps/api/v1/views.py]: Ok nice, i didn't understand this at first, but [this article](https://medium.com/@stevelukis/using-listapiview-unit-test-in-django-rest-framework-5af231ce246) helped me understand it

File: edx_exams/apps/api/v1/tests/test_views.py [hunk 7]
```diff
@@ -256,3 +256,45 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+
+        # import pudb; pu.db
+
+        response = self.get_response()
+        proctoring_providers_list = ProctoringProvider.objects.all()
+
+        expected_data = [self.test_provider.verbose_name, test_provider2.verbose_name]
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(proctoring_providers_list), 2)
```
ashultz0, line 289: this line can go, it's just testing the objects.all

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,45 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+        test_provider2 = ProctoringProvider.objects.create(
+            name='test_provider2',
+            verbose_name='testing_provider2',
+            lti_configuration_id='223456789'
+        )
+
+        # import pudb; pu.db
+
+        response = self.get_response()
+        proctoring_providers_list = ProctoringProvider.objects.all()
```
ashultz0, line 284: don't need to pull this

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,45 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+
+        expected_data = [self.test_provider.verbose_name, test_provider2.verbose_name]
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(proctoring_providers_list), 2)
+        self.assertIn(response.data[0]['verbose_name'], expected_data)
+        self.assertIn(response.data[1]['verbose_name'], expected_data)
```
ashultz0, line 291: this would pass if response.data[0] and response.data[1] are the same provider repeated twice by accident

one option is to sort the expected data (maybe call it expected names?) and then sort these two names, and then compare the sorted lists

another option would be to remove the name[0] from the expected data and then check name[1] against the remaining single item list

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +180,20 @@ def patch(self, request, course_id):
...
         return Response(status=response_status, data=data)
+
+
+class ProctoringProvidersView(ListAPIView):
+    """
+    Retrieve a list of all available proctoring providers
+
+    This endpoint returns an object for each provider
+
+    Path: /api/[version]/intents/most-recent-and-certain
```
Zacharis278, line 191: looks like some copypasta from PIE

edx_exams/apps/api/v1/views.py
```diff
@@ -179,3 +180,20 @@ def patch(self, request, course_id):
...
+    Returns:
+     * 200: OK, list of ProctoringProviderObjects
+    """
+
+    authentication_classes = (JwtAuthentication,)
```
MichaelRoytman, line 196: So neat! 🎉

edx_exams/apps/api/serializers.py
```diff
@@ -41,3 +41,13 @@ def validate_exam_type(self, value):
         if value not in valid_exam_types:
             raise serializers.ValidationError("Must be a valid exam type.")
         return value
+
+
+class ProctoringProviderSerializer(serializers.ModelSerializer):
```
MichaelRoytman, line 46: I was thinking a little bit about this. Is there a chance we're going to add any fields to the `ProctoringProvider` model that we _don't_ want to be publicly accessible? I don't think we've figured out where or how we're storing configuration data for providers. If so, we'll want to explicitly call out what fields we want to be returned by the API in this serializer instead of "all" so we don't inadvertently leak sensitive data in the future.

CC: @Zacharis278

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -256,3 +256,45 @@ def test_exams_config(self, other_course_id, exam_type, expect_none_provider):
...
+        return response
+
+    def test_proctoring_providers_list(self):
+        test_provider2 = ProctoringProvider.objects.create(
+            name='test_provider2',
+            verbose_name='testing_provider2',
+            lti_configuration_id='223456789'
+        )
+
+        # import pudb; pu.db
```
MichaelRoytman, line 281: I do this all the time, but don't forget to remove this line.

edx_exams/apps/api/serializers.py
Zacharis278: I don't foresee anything in this table being sensitive since this is mostly just an intermediate between an exam and an LTIConfiguration (which would be sensitive).  That's said I think you have a point here, and it's overall safer to enumerate the fields we're returning even if it is all of them.

Could save an issue of my assumption changes in the future

edx_exams/apps/api/v1/views.py
m-pitera: oops, it totally is haha

edx_exams/apps/api/v1/views.py
m-pitera: Thanks for the tip! I didn't realize how simple it'd be haha 😅

edx_exams/apps/api/serializers.py
m-pitera: got it, this makes sense

edx_exams/apps/api/v1/tests/test_views.py
m-pitera: sounds good, with the sort implementation i took an extra step to filter the namessy into a separate list to avoid dealing with OrderedDicts.

Zacharis278:
@m-pitera is this ready for another round of review?

m-pitera:
> @m-pitera is this ready for another round of review?

Yessir, sorry for delay was out for lunch 😅

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -346,3 +346,46 @@ def test_config_create(self):
...
+        )
+        response = self.get_response()
+        response_data_list = []
+
+        for item in response.data:
+            response_data_list.append(item['name'])
+
+        response_data_list.sort()
+
+        expected_data = sorted([self.test_provider.name, test_provider2.name])
```
ashultz0, line 378: this variable just gets in the way now, you could drop the two names right into the assert below

Zacharis278 (approved):
this looks great! 🚀

PR #35, opened 2022-09-01
MST-1525: exam forwarding middleware

Zacharis278:
**JIRA:** [MST-1525](https://2u-internal.atlassian.net/browse/MST-1525?atlOrigin=eyJpIjoiZGQyYTdhNWZkZGY5NDNlMjg0M2U2MGUwYjFmZDhhMjUiLCJwIjoiaiJ9)

**Description:**

1. Introduces router app to intercept requests and forward requests to edx-proctoring based on the courses configured proctoring provider
2. Allows a courses proctoring provider to be set to null. This is needed in the event we need to switch a course back from and LTI provider to one configured in edx-proctoring.

**Author concerns:** This introduces some unfortunate weirdness with timed exams. A course needs to universally use one exam service or the other and we've chosen to look at the configured proctoring provider to make that determination. That means timed exams are also impacted by selecting an LTI provider. We might need to revisit this if it becomes confusing/cumbersome to roll out.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**


edx_exams/apps/api/serializers.py
```diff
@@ -16,7 +16,7 @@ class ExamSerializer(serializers.ModelSerializer):
     course_id = serializers.CharField(required=False)
     content_id = serializers.CharField(required=True)
     time_limit_mins = serializers.IntegerField(required=True)
-    due_date = serializers.DateTimeField(required=False, format=None)
+    due_date = serializers.DateTimeField(required=True, format=None, allow_null=True)
```
Zacharis278, line 19: the view will blow up if this property doesn't exist so I think changing this to be required but allow null is prudent.

edx_exams/apps/router/interop.py
```diff
@@ -0,0 +1,21 @@
...
+Module for writing data to edx-proctoring
+"""
+from posixpath import join as urljoin
+
+from django.conf import settings
+from requests.exceptions import HTTPError
+
+from edx_exams.apps.core.rest_utils import get_client, make_request
+
+LMS_REGISTER_PROCTORED_EXAMS_API_TPL = 'api/edx_proctoring/v1/proctored_exam/exam_registration/course_id/{}'
```
ashultz0, line 11: yay searchable

edx_exams/apps/router/interop.py
```diff
@@ -0,0 +1,21 @@
...
+
+def register_exams(course_id, request_data):
```
ashultz0, line 13: request data is a list of exams, maybe call it exams or exam_list or something

or just quickly go add types to everything in edx-exams ;)

edx_exams/apps/router/middleware.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Intercept requests and determine if exams for this course
+    should be managed by legacy edx-proctoring system. If so, requests
+    are intercepted and forwarded to that service
+    """
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        try:
+            override_view = OVERRIDE_VIEW_MAP.get(view_func.view_class)
+        except AttributeError:
+            override_view = None
```
ashultz0, line 25: this feels like it would make more sense in form of an if
if func in map
view = view
else view = none

because it's not really an error, it's just not in the map

ashultz0 (commented):
some minor stuff for today, will try to read more before vacation but may not get to it

looks generally fine

edx_exams/apps/router/middleware.py
```diff
@@ -0,0 +1,32 @@
+"""
+Middleware that checks if in incoming request should be
+routed to the legacy proctoring service (edx-proctoring)
```
ashultz0, line 3: maybe this class and the map and so on should be named legacy or proctoring? Since this is not a generic override, it's the ExamRequestLegacyRedirectMiddleware. Except that's too long.

edx_exams/apps/router/middleware.py
```diff
@@ -0,0 +1,32 @@
...
+    are intercepted and forwarded to that service
+    """
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        try:
+            override_view = OVERRIDE_VIEW_MAP.get(view_func.view_class)
+        except AttributeError:
+            override_view = None
+
+        # call into override if a function is defined for this request method
+        if override_view and getattr(override_view, request.method.lower()):
```
ashultz0, line 28: the getattr is always a bad sign as is the fact that you could invisibly have a POST override but no PATCH override

Would it make sense to require that the override views cover the same methods as the originals? And just let things fail messily if they don't.

Or make the map be method specific?
(CourseExamsView, PATCH): CourseExamsOverrideView
after writing that out I really don't like it

Zacharis278: yeah I was trying to have this fall through seamlessly but now that you point it out that's pretty undesirable. I can see no reason we would ever want GET to go one place and PATCH to go another 😬

edx_exams/apps/router/middleware.py
Zacharis278: the reason I have this here is some functions passed to this middleware may not have a view_class attribute so this simplifies having to check if that attribute exists first.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -219,10 +219,8 @@ def test_exam_modified_type(self):
         self.assertEqual(proctored_exam.is_active, True)
 
     @ddt.data(
-        (True, 'proctored', True),  # test case for a proctored exam with no course config
         (False, 'proctored', False),  # test case for a proctored exam with a course config
         (False, 'timed', True),  # test case for a timed exam with a course config
-        (True, 'timed', True)  # test case for a timed exam with no course config
```
Zacharis278, line 225: at the moment, if there's no course config these requests are redirected to edx-proctoring.

varshamenon4: Question: could you clarify what the first True in the tuple signifies? Thanks!

edx_exams/apps/api/v1/views.py
```diff
@@ -123,7 +123,7 @@ def handle_exams(cls, request_exams_list, course_exams_qs, course_id):
                 provider = None
                 # get exam type class, which has specific attributes like is_proctored, is_timed, etc.
                 exam_type_class = get_exam_type(exam['exam_type'])
-                config = CourseExamConfiguration.objects.filter(course_id=course_id).first()
+                config = CourseExamConfiguration.get_configuration_for_course(course_id)
```
varshamenon4, line 126: Question: I see that this approach feels more readable (i.e. using `get_configuration_for_course` method instead of `filter`), but is there any other reason to use this approach rather than filter? Is filter more prone to bugginess here? Just curious. :)

edx_exams/apps/api/v1/views.py
```diff
@@ -212,26 +212,34 @@ def handle_config(cls, provider, course_id):
...
         """
+        error = None
+
         # check that proctoring provider is in request
-        if request.data.get('provider') is None:
-            response_status = status.HTTP_400_BAD_REQUEST
-            data = {"detail": "No proctoring provider name in request."}
+        if 'provider' not in request.data:
+            error = {"detail": "No proctoring provider name in request."}
+        elif request.data.get('provider') is None:
```
varshamenon4, line 227: Question: could you clarify the difference between the if and elif conditions? This basically allows for the provider to be 'None' in the case described in the PR description if the provider needs to be switched back from LTI to proctoring?

edx_exams/apps/api/v1/views.py
```diff
@@ -212,26 +212,34 @@ def handle_config(cls, provider, course_id):
...
-                self.handle_config(provider, course_id)
-                response_status = status.HTTP_204_NO_CONTENT
-                data = {}
             # return 400 if proctoring provider does not exist
             except ObjectDoesNotExist:
-                response_status = status.HTTP_400_BAD_REQUEST
-                data = {"detail": "Proctoring provider does not exist."}
+                error = {"detail": "Proctoring provider does not exist."}
+
+        if not error:
```
varshamenon4, line 236: Nice, this logic/flow makes sense to me!

edx_exams/apps/core/exam_types.py
```diff
@@ -35,9 +35,35 @@ class TimedExamType(ExamType):
...
+    Properties for practice exam
+    """
+    name = 'practice'
+    description = 'Practice, non-proctored, timed exam'
+    is_proctored = False
+    is_timed = True
+    is_practice = True
+
+
+class OnboardingExamType(ExamType):
```
varshamenon4, line 49: Suggestion: when this is eventually changed, would this need to be changed? Are there any other changes? I like that you included comments on this, but wondering if any additional documentation or a different approach would make the change cleaner from not supporting onboarding exams to supporting.

edx_exams/apps/core/migrations/0008_alter_exam_exam_type.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('core', '0007_alter_proctoringprovider_name'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='exam',
+            name='exam_type',
+            field=models.CharField(choices=[('proctored', 'Non-practice, timed, proctored exam'), ('timed', 'Non-practice, non-proctored, timed exam'), ('practice', 'Practice, non-proctored, timed exam'), ('onboarding', 'Practice, timed, proctored exam')], db_index=True, max_length=255),
```
varshamenon4, line 16: Nit: For some of the choices, timed comes before proctored and for others after:
'Non-practice, timed, proctored exam'
'Non-practice, non-proctored, timed exam'
For readability, you could consider having the order be consistent for these strings.

edx_exams/apps/core/models.py
```diff
@@ -142,11 +142,22 @@ class CourseExamConfiguration(TimeStampedModel):
...
 
     allow_opt_out = models.BooleanField(default=False)
 
     class Meta:
         """ Meta class for this Django model """
         db_table = 'exams_courseexamconfiguration'
         verbose_name = 'course exam configuration'
+
+    @classmethod
+    def get_configuration_for_course(cls, course_id):
```
varshamenon4, line 155: Nice!

edx_exams/apps/core/rest_utils.py
```diff
@@ -0,0 +1,37 @@
+"""
+Utils for reading and writing data to other services using REST.
+"""
+from django.conf import settings
+from edx_rest_api_client import client as rest_client
+
+
+# pylint: disable=inconsistent-return-statements
+def make_request(method, url, client, **kwargs):
```
varshamenon4, line 9: Question: is there a reason that a method needs to be created for this? As in, is there not an existing method that does this? Is there something custom happening here? Thanks for clarifying!

edx_exams/apps/core/rest_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Helper method to make an http request using
+    an authN'd client.
+    """
+    if method not in ['GET', 'POST', 'PUT', 'PATCH', 'DELETE']:  # pragma: no cover
```
varshamenon4, line 14: Question: what is "pragma: no cover"?

edx_exams/apps/router/middleware.py
```diff
@@ -0,0 +1,35 @@
...
+Middleware that checks if in incoming request should be
+routed to the legacy proctoring service (edx-proctoring)
+"""
+from django.utils.deprecation import MiddlewareMixin
+
+from edx_exams.apps.api.v1.views import CourseExamsView
+from edx_exams.apps.core.models import CourseExamConfiguration
+from edx_exams.apps.router.views import CourseExamsLegacyView
+
+LEGACY_VIEW_MAP = {
```
varshamenon4, line 11: Question: Is this a map because we anticipate multiple views having legacy views to be mapped to?

varshamenon4 (commented):
I've gone through most of the code and it looks great! Most of the comments are just clarifying questions for my own learning. A couple of suggestions, too. I still need to finish the last bit, but I wanted to get out these comments now before our meeting!

edx_exams/apps/api/v1/tests/test_views.py
Zacharis278: it signifies using a different course id that doesn't have a configuration in edx-exams so I removed these cases for the time being since so long as the new router app exists those requests will never hit this view

edx_exams/apps/api/v1/views.py
Zacharis278: mostly just about creating a reusable function for a commonly used database operations. As far as using get() under the hood instead of filter() the query will only ever return zero or one objects since course_id is unique at the database level. IMO it's more appropriate then returning the first element in a list of 1

edx_exams/apps/api/v1/views.py
Zacharis278: this ensures you have to explicitly set the provider to `null` in the request. So the first check makes sure 'provider' is afield in the request and the second checks if the value itself is None/null.

Actually this reminds me I need to throw a test in here 💡

edx_exams/apps/router/middleware.py
Zacharis278: yup 👍

edx_exams/apps/router/tests/test_interop.py
```diff
@@ -0,0 +1,60 @@
...
+        return inner
+
+    @ddt.data(200, 422)
+    @mock_oauth_login
+    @responses.activate
+    def test_register_exams(self, response_status):
+        """
+        Request is authenticated and forwarded to the LMS.
+        HTTP exceptions are handled and response is returned for
+        non-200 states codes
```
varshamenon4, line 50: Question: does this test both 200 and non-200 status codes?

edx_exams/apps/router/tests/test_views.py
```diff
@@ -0,0 +1,211 @@
...
+                },
+            ]
+        )
+        self.assertEqual(Exam.objects.all().count(), 1)
+
+    @mock.patch('edx_exams.apps.router.views.register_exams')
+    def test_patch_exams_empty(self, mock_register_exams):
+        """
+        Request is forwarded to the LMS, no exams in this
+        service should be removed.
```
varshamenon4, line 130: Question: could you clarify why exams could/would be removed? In other words, what is being tested here? Thanks!

varshamenon4 (commented):
Looks great!

edx_exams/apps/api/v1/views.py
varshamenon4: That makes sense to me! thanks for clarifying.

edx_exams/apps/router/views.py
```diff
@@ -0,0 +1,52 @@
+"""
+Views to wrap interactions with the legacy procuring service
```
schenedx, line 2: mis-spelled `proctoring`. Also, please reference what the legacy proctoring service referring to: github.com/openedx/edx-proctoring repo.

edx_exams/apps/router/interop.py
```diff
@@ -0,0 +1,25 @@
...
+from edx_exams.apps.core.rest_utils import get_client, make_request
+
+LMS_REGISTER_PROCTORED_EXAMS_API_TPL = 'api/edx_proctoring/v1/proctored_exam/exam_registration/course_id/{}'
+
+
+def register_exams(course_id, exam_list):
+    """
+    Register a list of course exams with the legacy proctoring service
+    """
+    path = LMS_REGISTER_PROCTORED_EXAMS_API_TPL.format(course_id)
```
schenedx, line 18: Suggestion, refactor the URL construction into it's own function so it can be reused later. Make the URL path template a bit more generic would be nice too.

schenedx (approved):
2 quick and NIT suggestions. LGTM 👍 Good work

edx_exams/apps/core/rest_utils.py
Zacharis278: not too much going on here really it's just to standardize calling `OAuthAPIClient.request()` and keeping that code alongside the `OAuthAPIClient` initialization. Also this throws an exception for non 200 responses so implementers don't have to check for that.

edx_exams/apps/router/tests/test_interop.py
Zacharis278: yeah with 🪄 , the `ddt.data` decorator will generate two copies of this test, one for 200 and one for 422.  When you run these they output something along the lines of `test_register_exams_1_200`, `test_register_exams_2_422`

edx_exams/apps/router/tests/test_views.py
Zacharis278: updated the comment a bit but the patch endpoint takes all of the exams in the course and syncs to that. So if studio sends an empty list, it means all exams would be removed.

edx_exams/apps/core/exam_types.py
Zacharis278: I don't think any thing would need to change here, if we did create an exam of this type it would 'just work' as proctored but we aren't expecting to be setting these up initially or building onboarding specific features yet.

varshamenon4: That makes sense! Thanks for clarifying.

edx_exams/apps/router/tests/test_interop.py
varshamenon4: Well, that's cool! Thanks!

varshamenon4 (approved):
Looks good to me! Thanks for answering my questions. :)

PR #38, opened 2022-09-23
feat: drop anonymous_user_id column for user

varshamenon4:
**JIRA:** [MST-1644](https://2u-internal.atlassian.net/browse/MST-1644)

**Description:** This PR drops the anonymous_user_id column for User model. Includes migration. Done as one release since there are no usages of the field. The re-adding of the anonymous_user_id as a UUIDField will be a separate PR/release.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/core/models.py
+++ b/edx_exams/apps/core/models.py
@@ -20,8 +20,6 @@ class User(AbstractUser):
     full_name = models.CharField(_('Full Name'), max_length=255, blank=True, null=True)
     lms_user_id = models.IntegerField(null=True, db_index=True)
 
-    anonymous_user_id = models.IntegerField(null=True, db_index=True)
-
     @property
     def access_token(self):
         """


edx_exams/apps/core/models.py
```diff
@@ -9,6 +10,8 @@
 
 
 class User(AbstractUser):
+    # todo: should I update this docstring to remove username and email address and change to
```
varshamenon4, line 13: I wasn't sure if this docstring is meant to include username and email address or if this is outdated. If so, I can update to reflect current user model? I do see that these fields are in the AbstractUser so maybe this is why it's here as PII?

edx_exams/apps/core/models.py
```diff
@@ -18,9 +21,10 @@ class User(AbstractUser):
 
     """
     full_name = models.CharField(_('Full Name'), max_length=255, blank=True, null=True)
+
     lms_user_id = models.IntegerField(null=True, db_index=True)
 
-    anonymous_user_id = models.IntegerField(null=True, db_index=True)
+    anonymous_user_id = models.UUIDField(default=uuid.uuid4, db_index=True, unique=True)
```
varshamenon4, line 27: Should I add any tests?

varshamenon4 (commented):
Just some clarifying questions about doc string and testing. Thanks for the help!

edx_exams/apps/core/models.py
ashultz0: this docstring is all PII retirement related so I don't think there is anything to add from this work, anonymous ID is not really PII

edx_exams/apps/core/models.py
ashultz0: the only thing to test really is that django's field definition works which isn't worth testing

ashultz0 (changes requested):
this is all theoretically correct but the mechanics of how migrations actually happen make it dangerous

https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations is the go to source here. The short version is the migration happens before the code update.

So for a while we'll have database as UUID, code as int. This means any user access may explode.

Additionally in the DB itself if there are any columns with data instead of just nulls turning an int into a UUID probably doesn't work. If we are copying this database anywhere like snowflake, that would also need to change type. (I don't think we are in this case)

Since this is a new unused service it's possible that neither of these will actually cause a problem. But it also means it's a free chance to learn safe migrations in a low risk environment.

edx_exams/apps/core/models.py
```diff
@@ -20,8 +20,6 @@ class User(AbstractUser):
     full_name = models.CharField(_('Full Name'), max_length=255, blank=True, null=True)
     lms_user_id = models.IntegerField(null=True, db_index=True)
 
-    anonymous_user_id = models.IntegerField(null=True, db_index=True)
```
varshamenon4, line 23: @ashultz0 per your recommendation (thanks for sharing that doc), my plan is to separate this into two parts. 

1. In this PR, I'm dropping the anonymous_user_id int column based on the steps here: [Everything About Database Migrations](https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations).
2. Then, I will follow the steps in the doc to add a new anonymous_user_id UUIDField with the default as a UUID. 

I decided to do it this way than doing an AlterField, as this seemed to be a little less safe and also we don't need the previous anonymous_user_id int field so there's no operation required here to copy over the values. Does this make sense/sound good?

varshamenon4 (commented):
Shifted the approach to make it a safer migration. Thanks!

edx_exams/apps/core/models.py
ashultz0: I think that would be my first instinct - just kill the old one and put a new one back in the same place

ashultz0: but per the DB migrations doc I think you need to split the field removal and the column removal migration into two separate tiny releases

varshamenon4: Ah that's right! I misread that--I'll update!

ashultz0 (approved):

PR #44, opened 2022-10-04
fix: update dictionary settings list

alangsto:
**Description:** The sandbox config was overriding `EDX_DRF_EXTENSIONS`, which prevents user attributes from being correctly mapped from the JWT token. There is already a constant in the production settings file used to keep track of dictionary settings (so that only keys within the dictionary are updated, not the entire dictionary), so `EDX_DRF_EXTENSIONS` should be added to that constant.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**


schenedx (approved):
👍

PR #45, opened 2022-10-05
feat: generate anonymous_user_id field in User model

varshamenon4:
**JIRA:** [MST-1644](https://2u-internal.atlassian.net/browse/MST-1644)

**Description:** This PR is part of the work above to generate the anonymous_user_id field as a UUID in the User model. This approach adds the field, then does the operation to update it, then updates the field.

**Author concerns:** This migration uses a similar approach as [this one](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/migrations/0008_fix_uuid_backfill.py) in order to create unique UUIDs. I want to confirm that this is the correct approach.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

**Post merge:**


Zacharis278 (approved):
makes sense to me. Were you able to test applying this migration locally on a DB without the id field?

PR #47, opened 2022-10-07
feat: get endpoint for exam config

Zacharis278:
**JIRA:** [MST-1527](https://2u-internal.atlassian.net/browse/MST-1527)

**Description:** GET course exam config. This is needed to fetch the currently selected provider in course-authoring.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/api/v1/views.py
+++ b/edx_exams/apps/api/v1/views.py
@@ -12,7 +12,7 @@ from rest_framework.generics import ListAPIView
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
-from edx_exams.apps.api.permissions import StaffUserPermissions, StaffUserOrReadOnlyPermissions
+from edx_exams.apps.api.permissions import StaffUserOrReadOnlyPermissions, StaffUserPermissions
 from edx_exams.apps.api.serializers import ExamSerializer, ProctoringProviderSerializer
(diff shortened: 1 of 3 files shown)


edx_exams/apps/api/v1/views.py
```diff
@@ -215,6 +215,19 @@ def handle_config(cls, provider, course_id):
         provider_name = provider.name if provider else None
         log.info(f"Created or updated course exam configuration course_id={course_id},provider={provider_name}")
 
+    def get(self, request, course_id):
```
Zacharis278, line 218: we should probably make a serializer for this if more fields get added but I chose not to go down that path since this was simpler and does't require altering the PATCH function.

ashultz0: Do we expect they will stay in sync or get out of sync? If we want to make sure they don't get out of sync maybe a quick comment in each about "keep these in sync or make a serializer"

I can see patch taking more things than we are willing to provide to get. Although I guess that means they are really two separate endpoints rather than the same one.

Zacharis278: IMO these should stay in sync if they're at the same URL. I'll add a comment to note this for future us

PR #51, opened 2022-10-27
docs: testing plan for exam configuration

Zacharis278:
**JIRA:** [MST-1657](https://2u-internal.atlassian.net/browse/MST-1657)

**Description:** Document testing steps for exam setup and provider configuration. (Dev milestone 2)

The plan is to have a separate file for each milestone/feature set to keep things organized. That is, assuming we choose to continue this approach in future milestones.

### DIFF
--- a/docs/test_plan/exam_configuration.rst
+++ b/docs/test_plan/exam_configuration.rst
@@ -1,8 +1,14 @@
 Exam Setup and Configuration
 ============================
 
-Exam Configuration (edx-exams)
-------------------------------
+Helpful ReadTheDocs Pages:
+
+- `Configuring a Proctoring Provider <https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/proctored_exams/proctored_enabling.html#configuring-proctoring-provider>`_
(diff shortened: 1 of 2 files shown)


docs/test_plan/exam_configuration.rst
```diff
@@ -0,0 +1,189 @@
+Exam Setup and Configuration
+============================
+
+Exam Configuration (edx-exams)
+------------------------------
+
+Prerequisite Steps:
+   #. Make sure `course_apps.exams_ida` CourseWaffleFlag is enabled for your test course
+   #. An LTI proctoring provider should be chosen for the course
```
ashultz0, line 9: we're presumably using a fake one? can we put setup instructions in a section here, or maybe in another doc?

Zacharis278: humm yeah maybe I can lean on the read the docs pages more here and elsewhere in this doc

schenedx: Agree. As this is written, I won't know how to do this

schenedx (approved):
👍  Regardless of my comment, this document is very useful for anyone who reads and use the code.

PR #53, opened 2022-11-10
feat: when updating course exam config, rebuild exams if new provider

ashultz0:
It turns out to be very easy to churn out duplicates in django.

A few callouts:
1. Since we build a new set of duplicate exams and can flipflop back to a previous provider, we have to remove the existing uniqueness constraint on exams.
2. The original ticket specifies that only staff can make this change if the exams have attempts. As the endpoint is currently set up, only staff can do writes at all so no extra code is needed.
3. The whole "this model's update changes other models" felt like a model concern so I moved all that code out of the view.  I thought about moving the testing as well but it is set up pretty well where it is so I left it.

**JIRA:** [Link to JIRA ticket](https://2u-internal.atlassian.net/browse/MST-1579)


edx_exams/apps/core/models.py
```diff
@@ -163,3 +165,51 @@ def get_configuration_for_course(cls, course_id):
         except cls.DoesNotExist:
             configuration = None
         return configuration
+
+    @classmethod
+    @transaction.atomic
+    def create_or_update(cls, provider, course_id):
```
Zacharis278, line 171: agreed that this makes more sense here

Zacharis278:
we'll need something to determine role if we make provider configuration and setup more self-service, maybe there's a better pattern out there?

Zacharis278 (approved):
👍

PR #59, opened 2022-12-15
feat: authenticate LTI exams launch

Zacharis278:
**JIRA:** [MST-1724](https://2u-internal.atlassian.net/browse/MST-1724?atlOrigin=eyJpIjoiNzliZTFjYjA1YzI1NGRlMTlhYThkODU2ZWI5YmY2NDgiLCJwIjoiaiJ9)

**Description:** 
In order to make use of the edx browser JWT the client needs to explicitly request its use with a custom header. As a result, any redirect that occurs as part of the LTI launch flow cannot be authenticated using this JWT. This PR will create an explicit django session with the exam service when `/start_proctoring/` is called so those LTI callbacks will authenticate using the django session cookie. This is the one endpoint that can make use of the JWT since we expect this request to come from the learning MFE.

The downside for now is the initial request must come from the MFE and simply doing a GET request to /start_proctoring/<id> will no longer work. If you want to do a launch you'll need to set the use-jwt header using a browser tool. Here's how I'm doing that with requestly https://app.requestly.io/rules/#sharedList/1671117948648-apply-edx-exams-jwt

### DIFF
--- a/edx_exams/apps/lti/utils.py
+++ b/edx_exams/apps/lti/utils.py
@@ -5,7 +5,4 @@ from django.conf import settings
 
 
 def get_lti_root():
-    if hasattr(settings, 'LTI_ROOT_URL_OVERRIDE'):
-        return settings.LTI_ROOT_URL_OVERRIDE
-    else:
-        return settings.ROOT_URL
+    return settings.ROOT_URL
--- a/edx_exams/apps/lti/views.py
+++ b/edx_exams/apps/lti/views.py
@@ -4,10 +4,14 @@ LTI Views
 
 from urllib.parse import urljoin
 
+from django.contrib.auth import login
 from django.http import JsonResponse
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.views.decorators.http import require_http_methods
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+from rest_framework.decorators import authentication_classes, permission_classes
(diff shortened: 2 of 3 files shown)


edx_exams/apps/lti/utils.py
```diff
@@ -5,7 +5,4 @@
 
 
 def get_lti_root():
-    if hasattr(settings, 'LTI_ROOT_URL_OVERRIDE'):
```
Zacharis278, line 8: we won't ever need a separate url for this anymore based on this setup: https://2u-internal.atlassian.net/wiki/spaces/PT/pages/256737327/Local+Development+LTI+Configuration

edx_exams/apps/lti/views.py
```diff
@@ -4,26 +4,39 @@
...
     """
     # TODO: Here we'd do all the start of proctoring things.
 
-    exam_attempt = ExamAttempt.objects.get(pk=attempt_id)
+    # user is authenticated via JWT so use that to create a
+    # session with this service's authentication backend
+    request.user.backend = EDX_OAUTH_BACKEND
+    login(request, request.user)
+
+    exam_attempt = ExamAttempt.objects.get(pk=attempt_id, user=request.user)
```
Zacharis278, line 39: added this very basic check here to prevent accidentally messing with another user's test attempt.

ashultz0 (approved):

PR #61, opened 2022-12-22
feat: add exam access ticket endpoint

varshamenon4:
**JIRA:** [MST-1598](https://2u-internal.atlassian.net/browse/MST-1598)

**Description:** This ticket adds a new http endpoint to edx-exams that gets exam access tickets.

### DIFF
--- a/edx_exams/apps/api/utils.py
+++ b/edx_exams/apps/api/utils.py
@@ -1,0 +1,34 @@
+"""
+API Utility Functions
+"""
+import logging
+from datetime import timedelta
+
+from django.utils import timezone
+
+log = logging.getLogger(__name__)
+
+
+def get_exam_attempt_time_remaining(exam_attempt, now=None):
+    """
+    For timed exam attempt, get the time remaining in seconds.
+
+    If the end time has already passed, set time remaining to zero.
+    """
+    if exam_attempt.start_time is None or exam_attempt.allowed_time_limit_mins is None:
+        return 0
+
+    now = now or timezone.now()
+
+    # Calculate end of attempt time
+    end_time = exam_attempt.start_time + timedelta(minutes=exam_attempt.allowed_time_limit_mins)
+
+    # Compare end of attempt time to now
+    # If end time has passed, set to zero
+    if end_time > now:
+        secs_to_end = 0
+    # Else calculate seconds remaining for attempt
+    else:
+        secs_to_end = (now - end_time).total_seconds()
+
+    return secs_to_end
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -3,15 +3,17 @@ Tests for the exams API views
 """
 import json
 import uuid
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import ddt
 import pytz
 from django.urls import reverse
+from django.utils import timezone
+from token_utils.api import unpack_token_for
 
 from edx_exams.apps.api.test_utils import ExamsAPITestCase
 from edx_exams.apps.api.test_utils.factories import UserFactory
-from edx_exams.apps.core.models import CourseExamConfiguration, Exam, ProctoringProvider
+from edx_exams.apps.core.models import CourseExamConfiguration, Exam, ExamAttempt, ProctoringProvider
 
 
 @ddt.ddt
@@ -513,3 +515,265 @@ class ProctoringProvidersViewTest(ExamsAPITestCase):
 
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.data), 0)
+
+
+@ddt.ddt
+class ExamAccessTicketsViewsTests(ExamsAPITestCase):
+    """
+    Tests for Exam Access Ticket Views.
+    """
+
+    def setUp(self):
+        super().setUp()
+
+        self.course_id = 'course-v1:edx+test+f19'
+
+        self.due_date = timezone.now() + timedelta(minutes=5)
+        self.exam = Exam.objects.create(
+            resource_id=str(uuid.uuid4()),
+            course_id=self.course_id,
+            provider=self.test_provider,
+            content_id='abcd1234',
+            exam_name='test_exam',
+            exam_type='proctored',
+            time_limit_mins=30,
+            due_date=self.due_date,
+            hide_after_due=False,
+            is_active=True
+        )
+        self.exam_id = self.exam.id
+        self.url = reverse('api:v1:exam-access-tickets', kwargs={'exam_id': self.exam_id})
+
+        self.past_due_date = timezone.now() - timedelta(minutes=5)
+        self.past_due_exam = Exam.objects.create(
+            resource_id=str(uuid.uuid4()),
+            course_id=self.course_id,
+            provider=self.test_provider,
+            content_id='abcd1234',
+            exam_name='test_exam',
+            exam_type='proctored',
+            time_limit_mins=30,
+            due_date=self.past_due_date,
+            hide_after_due=False,
+            is_active=True
+        )
+
+        self.past_due_exam_id = self.past_due_exam.id
+        self.past_due_url = reverse('api:v1:exam-access-tickets', kwargs={'exam_id': self.past_due_exam_id})
+
+    def get_exam_access(self, user, url):
+        """
+        Helper function to make a get request
+        """
+        headers = self.build_jwt_headers(user)
+        return self.client.get(url, **headers)
+
+    def assert_valid_exam_access_ticket(self, response, user, exam):
+        ticket = response.cookies["exam_access_ticket"].value
+        self.assertEqual(unpack_token_for(ticket, user.lms_user_id).get('course_id'), exam.course_id)
+        self.assertEqual(unpack_token_for(ticket, user.lms_user_id).get('content_id'), exam.content_id)
+
+    def test_auth_failures(self):
+        """
+        Verify the endpoint validates permissions
+        """
+        # Test unauthenticated
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 401)
+
+    def test_exam_not_found(self):
+        """
+        Verify the endpoint returns 404 if exam is not found
+        """
+        url = reverse('api:v1:exam-access-tickets', kwargs={'exam_id': 674})
+
+        headers = self.build_jwt_headers(self.user)
+        response = self.client.get(url, **headers)
+        self.assertEqual(response.status_code, 404)
+
+    def test_access_not_granted(self):
+        """
+        Verify the endpoint doesn't grant access for an exam
+        without an existing exam attempt or past due date.
+        """
+        response = self.get_exam_access(self.user, self.url)
+        self.assertEqual(403, response.status_code)
+
+    @ddt.data(
+        (False, 'started', 200),
+        (False, 'created', 403),
+        (True, 'verified', 200),
+        (True, 'created', 403),
+    )
+    @ddt.unpack
+    def test_access_granted_started_exam_attempt(self, exam_past_due, attempt_status, response_status):
+        """
+        Verify the endpoint grants/doesn't grant access for an exam
+        based on the status of the exam attempt.
+        """
+        exam = self.past_due_exam if exam_past_due else self.exam
+        due_date = self.past_due_date if exam_past_due else self.due_date
+
+        start_time = due_date - timedelta(minutes=60)
+        allowed_time_limit_mins = exam.time_limit_mins
+        ExamAttempt.objects.create(
+            user=self.user,
+            exam=exam,
+            attempt_number=1,
+            status=attempt_status,
+            start_time=start_time,
+            allowed_time_limit_mins=allowed_time_limit_mins
+        )
+
+        url = reverse('api:v1:exam-access-tickets', kwargs={'exam_id': exam.id})
+        response = self.get_exam_access(self.user, url)
+        self.assertEqual(response_status, response.status_code)
+        if response_status == 200:
+            self.assert_valid_exam_access_ticket(response, self.user, exam)
+
+    def test_access_not_granted_started_exam_attempt_missing_start_time(self):
+        """
+        Verify the endpoint does not grant access for an exam
+        with an existing, started exam attempt that is missing start_time.
+        """
+        allowed_time_limit_mins = self.exam.time_limit_mins
+        ExamAttempt.objects.create(
+            user=self.user,
+            exam=self.exam,
+            attempt_number=1,
+            status='started',
+            allowed_time_limit_mins=allowed_time_limit_mins
+        )
+
+        response = self.get_exam_access(self.user, self.url)
+        self.assertEqual(403, response.status_code)
+
+    @ddt.data(
+        ('started', 200),
+        ('created', 403),
+    )
+    @ddt.unpack
+    def test_access_no_due_date(self, attempt_status, response_status):
+        """
+        Verify the endpoint grants access for an exam
+        with no due date, if started exam attempt.
+        """
+        no_due_date_exam = Exam.objects.create(
+            resource_id=str(uuid.uuid4()),
+            course_id=self.course_id,
+            provider=self.test_provider,
+            content_id='abcd1234',
+            exam_name='test_exam',
+            exam_type='proctored',
+            time_limit_mins=30,
+            hide_after_due=False,
+            is_active=True
+        )
+
+        no_due_date_exam_id = no_due_date_exam.id
+        no_due_date_url = reverse('api:v1:exam-access-tickets', kwargs={'exam_id': no_due_date_exam_id})
+
+        start_time = timezone.now() - timedelta(minutes=60)
+        allowed_time_limit_mins = no_due_date_exam.time_limit_mins
+        ExamAttempt.objects.create(
+            user=self.user,
+            exam=no_due_date_exam,
+            attempt_number=1,
+            status=attempt_status,
+            start_time=start_time,
+            allowed_time_limit_mins=allowed_time_limit_mins
+        )
+
+        response = self.get_exam_access(self.user, no_due_date_url)
+        self.assertEqual(response_status, response.status_code)
+        if response_status == 200:
+            self.assert_valid_exam_access_ticket(response, self.user, no_due_date_exam)
+
+    def test_access_not_granted_if_hide_after_due(self):
+        """
+        Verify the endpoint does not grant access for past-due exam
+        when the exam is set to hide after due.
+        """
+        past_due_exam_hide = Exam.objects.create(
+            resource_id=str(uuid.uuid4()),
+            course_id=self.course_id,
+            provider=self.test_provider,
+            content_id='abcd1234',
+            exam_name='test_exam',
+            exam_type='proctored',
+            time_limit_mins=30,
+            due_date=self.past_due_date,
+            hide_after_due=True,
+            is_active=True
+        )
+
+        past_due_exam_id = past_due_exam_hide.id
+        past_due_url = reverse('api:v1:exam-access-tickets', kwargs={'exam_id': past_due_exam_id})
+
+        start_time = self.past_due_date - timedelta(minutes=60)
+        allowed_time_limit_mins = past_due_exam_hide.time_limit_mins
+        ExamAttempt.objects.create(
+            user=self.user,
+            exam=past_due_exam_hide,
+            attempt_number=1,
+            status='verified',
+            start_time=start_time,
+            allowed_time_limit_mins=allowed_time_limit_mins
+        )
+
+        response = self.get_exam_access(self.user, past_due_url)
+        self.assertEqual(403, response.status_code)
+
+    def test_access_not_granted_no_status_exam_attempt(self):
+        """
+        Verify the endpoint does not grant access for an exam
+        with an existing exam attempt that is not started.
+        """
+        start_time = self.due_date - timedelta(minutes=60)
+        allowed_time_limit_mins = self.exam.time_limit_mins
+        ExamAttempt.objects.create(
+            user=self.user,
+            exam=self.exam,
+            attempt_number=1,
+            start_time=start_time,
+            allowed_time_limit_mins=allowed_time_limit_mins
+        )
+
+        response = self.get_exam_access(self.user, self.url)
+        self.assertEqual(403, response.status_code)
+
+    @ddt.data(
+        (-timedelta(minutes=25), 403),  # exam attempt with zero time remaining
+        (timedelta(minutes=5), 403),  # exam attempt that is past due date
+        (-timedelta(seconds=2140), 200),  # exam attempt time remaining less than due date
+    )
+    @ddt.unpack
+    def test_access_granted_started_exam_attempt_various_times(self, start_time_delta, response_status):
+        """
+        Verify the endpoint grants access for an exam
+        with an existing exam attempt.
+        """
+        start_time = self.due_date + start_time_delta
+        allowed_time_limit_mins = self.exam.time_limit_mins
+        ExamAttempt.objects.create(
+            user=self.user,
+            exam=self.exam,
+            attempt_number=1,
+            status='started',
+            start_time=start_time,
+            allowed_time_limit_mins=allowed_time_limit_mins
+        )
+
+        response = self.get_exam_access(self.user, self.url)
+        self.assertEqual(response_status, response.status_code)
+        if response_status == 200:
+            self.assert_valid_exam_access_ticket(response, self.user, self.exam)
+
+    def test_access_granted_past_due_exam_no_attempt(self):
+        """
+        Verify the endpoint grants access for an exam
+        with no existing attempt that is past due.
+        """
+        response = self.get_exam_access(self.user, self.past_due_url)
+        self.assertEqual(200, response.status_code)
+        self.assert_valid_exam_access_ticket(response, self.user, self.past_due_exam)
--- a/edx_exams/apps/api/v1/views.py
+++ b/edx_exams/apps/api/v1/views.py
@@ -5,17 +5,22 @@ import logging
 import uuid
 
 from django.core.exceptions import ObjectDoesNotExist
+from django.utils import timezone
 from edx_api_doc_tools import path_parameter, schema
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from rest_framework import status
 from rest_framework.generics import ListAPIView
 from rest_framework.response import Response
 from rest_framework.views import APIView
+from token_utils.api import sign_token_for
 
 from edx_exams.apps.api.permissions import StaffUserOrReadOnlyPermissions, StaffUserPermissions
 from edx_exams.apps.api.serializers import ExamSerializer, ProctoringProviderSerializer
+from edx_exams.apps.api.utils import get_exam_attempt_time_remaining
 from edx_exams.apps.core.exam_types import get_exam_type
-from edx_exams.apps.core.models import CourseExamConfiguration, Exam, ProctoringProvider
+from edx_exams.apps.core.models import CourseExamConfiguration, Exam, ExamAttempt, ProctoringProvider
+from edx_exams.apps.core.statuses import ExamAttemptStatus
+from edx_exams.settings.base import ACCESS_TOKEN_COOKIE_DOMAIN, ACCESS_TOKEN_COOKIE_NAME
 
 log = logging.getLogger(__name__)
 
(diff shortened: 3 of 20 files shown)


edx_exams/apps/access/views.py
```diff
@@ -0,0 +1,63 @@
...
+        user = request.user
+        exam = request.exam
+        claims = {}
+        claims.update(exam.course_id)
+        claims.update(exam.content_id)
+        exp_secs = 60
+
+        exam_attempt = ExamAttempt.get_latest_user_exam_attempt(user, exam)
+        # todo: should this be called ticket or token? For consistency?
+        cookie_name = "exam_access_ticket"
```
varshamenon4, line 30: Question: should we call them exam_access_tickets? Or tokens? Just want to keep it consistent.

edx_exams/apps/access/views.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # If exam attempt exists for user, then grant exam access
+        if exam_attempt is not None:
+            # Grant access based on a short check-in window or the time to the end of the exam attempt,
+            # whichever is shorter.
+            if exam_attempt.start_time is not None and exam_attempt.allowed_time_limit_mins is not None:
```
varshamenon4, line 36: Question: does this logic make sense? I need to fix the datetime.now() on line 38 with some debugging (because of data type inconsistencies), but I just want to check that the logic overall is sound. Also curious if anyone has feedback on better ways to approach this.

edx_exams/apps/access/views.py
```diff
@@ -0,0 +1,63 @@
...
+                end_time = exam_attempt.start_time + timedelta(minutes=exam_attempt.allowed_time_limit_mins)
+                secs_to_end = (end_time - datetime.now()).total_seconds()
+                if secs_to_end < exp_secs:
+                    exp_secs = secs_to_end
+            access_ticket = create_jwt(user.lms_user_id, exp_secs, claims)
+            # todo: What should be included in data? Since cookie includes access ticket.
+            response = HttpResponse("Created exam access ticket for latest attempt",
```
varshamenon4, line 43: Question: since the cookie includes the access ticket, should anything go in data?

edx_exams/apps/access/views.py
```diff
@@ -0,0 +1,63 @@
...
+            if datetime.now() > exam.due_date:
+                access_ticket = create_jwt(user.lms_user_id, exp_secs, claims)
+                response = HttpResponse("Created exam access ticket for exam passed due date", 
+                                        status=status.HTTP_200_OK,
+                                        data=[])
+                response.set_cookie(cookie_name, access_ticket, max_age=exp_secs)
+
+        # Else, do not grant exam access ticket
+        else:
+            # todo: is this the correct status?
```
varshamenon4, line 59: Question: Is this the correct error status for not granting an exam access ticket? Or should there be no error?

edx_exams/apps/core/models.py
```diff
@@ -136,6 +137,14 @@ class Meta:
         db_table = 'exams_examattempt'
         verbose_name = 'exam attempt'
 
+    @classmethod
+    def get_latest_user_exam_attempt(cls, user, exam):
```
varshamenon4, line 141: Question: I'm just getting the latest exam attempt but do I need to consider the status or start_time of the exam attempt to get the most recent, current attempt?

varshamenon4 (commented):
Still need to do testing and add urls, etc, but wanted an initial review to address some specific questions! Thanks for the help.

edx_exams/apps/access/views.py
```diff
@@ -0,0 +1,63 @@
+"""
+Exam Access Ticket Views
```
varshamenon4, line 2: Question: I added an access package to add views/endpoints for access ticket specifically. Should this go somewhere else?

edx_exams/apps/access/views.py
```diff
@@ -0,0 +1,63 @@
...
+    # todo: should this be a get request?
+    def get(self, request):
+        user = request.user
+        exam = request.exam
+        claims = {}
+        claims.update(exam.course_id)
+        claims.update(exam.content_id)
+        exp_secs = 60
+
+        exam_attempt = ExamAttempt.get_latest_user_exam_attempt(user, exam)
```
Zacharis278, line 28: Most of the requests coming from the UI are exam attempt specific anyway. It might just be easier to take the attempt as an input instead of trying to look it up here.

edx_exams/apps/access/views.py
Zacharis278: I think we landed on using 'ticket' for this to avoid ambiguity with existing JWT 'tokens' so I'd go with that.

edx_exams/apps/access/views.py
Zacharis278: this whole block is logic we'll definitely need to reuse in other parts of this app. Maybe create a utility function for `get_attempt_time_remaining`? This should probably just return 0 if we're past the end time.

It looks like right now exp_secs is becoming negative if we're beyond the end time? How do you think we should handle that?

edx_exams/apps/access/views.py
```diff
@@ -0,0 +1,63 @@
...
+                end_time = exam_attempt.start_time + timedelta(minutes=exam_attempt.allowed_time_limit_mins)
+                secs_to_end = (end_time - datetime.now()).total_seconds()
+                if secs_to_end < exp_secs:
+                    exp_secs = secs_to_end
+            access_ticket = create_jwt(user.lms_user_id, exp_secs, claims)
+            # todo: What should be included in data? Since cookie includes access ticket.
+            response = HttpResponse("Created exam access ticket for latest attempt", 
+                                    status=status.HTTP_200_OK, 
+                                    data=[])
+            response.set_cookie(cookie_name, access_ticket, max_age=exp_secs)
```
Zacharis278, line 46: for the sake of troubleshooting I think we can leave the max_age parameter off this. The ticket includes it's own expiry so no need to purge it from the browser.

edx_exams/apps/access/views.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # If exam is passed the due date, then grant exam access
+        elif exam.due_date is not None:
+            if datetime.now() > exam.due_date:
+                access_ticket = create_jwt(user.lms_user_id, exp_secs, claims)
```
Zacharis278, line 51: these few lines are a repeat of the case where the exam is in progress. Can you refactor a bit so we only need to build the ticket/response one time?

edx_exams/apps/core/models.py
Zacharis278: latest created sounds about right. As a NIT I think this function might be clearer if we call it 'current' exam attempt since that's the language the UI uses for this.

user_id, and exam_id might be better input parameters here. Calls to this function are more likely to have the id's in scope than whole object.

edx_exams/apps/access/views.py
ashultz0: I think it was actually the other way around, I used "ticket" all over my confluence doc but people didn't like it hence token in the library name.

It doesn't actually matter but yes it should be consistent 😆

edx_exams/apps/access/views.py
ashultz0: probably better to add it later if so than guess now

edx_exams/apps/access/views.py
ashultz0: 403 is I think the correct code for "we recognize you as an authenticated person but you aren't allowed to see this". It's nicer to 403 and let the calling code know that immediately instead of secretly giving out no cookie

edx_exams/apps/access/views.py
```diff
@@ -0,0 +1,63 @@
...
+                response = HttpResponse("Created exam access ticket for exam passed due date", 
+                                        status=status.HTTP_200_OK,
+                                        data=[])
+                response.set_cookie(cookie_name, access_ticket, max_age=exp_secs)
```
ashultz0, line 55: I suspect we have to do something with the cookie to make sure it's set for the enclosing domain instead of specific domain

courses.edx.org needs to read the cookie
exams.edx.org is giving out the cookie, but if the cookie is set for exams.edx.org courses won't get it
exams needs to set it for just edx.org

this must be something we are already handing with the normal auth cookies, grep for COOKIE_DOMAIN or similar

edx_exams/apps/core/models.py
ashultz0: do we want to only get attempts that aren't yet expired?

edx_exams/apps/access/views.py
varshamenon4: Yeah, good call!

edx_exams/apps/access/views.py
varshamenon4: Sounds good.

edx_exams/apps/access/views.py
Zacharis278: I think you might be better off simply including this in the api app. Access is only going to ever include this one relatively small endpoint.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -513,3 +513,182 @@ def test_proctoring_providers_list_empty(self):
...
+            attempt_number=1,
+            status='started',
+            start_time=start_time,
+            allowed_time_limit_mins=allowed_time_limit_mins
+        )
+
+        response = self.get_exam_access(self.user)
+        self.assertEqual(200, response.status_code)
+        self.assertIsInstance(response.cookies["exam_access_ticket"].value, str)
+        # todo: also check that exp is correct in cookie?
```
varshamenon4, line 626: Question: Should I also check that the exp window is correct? If so, is it better to only test the helper method that calculates this (set_expiration_window) or would it be better to actually test that the expiration is correct in the cookie by decoding and checking?

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +274,116 @@ class ProctoringProvidersView(ListAPIView):
...
+        data = {}
+        grant_access = True
+        response_status = status.HTTP_200_OK
+
+        # If exam attempt exists for user, then grant exam access
+        if exam_attempt is not None:
+            log.info("Creating exam access ticket for attempt")
+            # If started, determine how to set expiration window
+            # todo: should exam attempt status be started or something else?
+            if exam_attempt.status == "started":
```
varshamenon4, line 349: Question: is this the correct check for the exam attempt status?

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +274,116 @@ class ProctoringProvidersView(ListAPIView):
...
+        return response
+
+    def get(self, request, exam_id):
+        """
+        Get exam access ticket as JWT added as cookie to response.
+
+        Exam access ticket corresponds to given exam and user in request.
+        """
+        # todo: Is try/except block necessary? Or does exam always exist?
+        try:
```
varshamenon4, line 380: Question: is a try/except needed here? Is there a case where the exam wouldn't exist in this case?

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +274,116 @@ class ProctoringProvidersView(ListAPIView):
...
+
+        403 error if access is not granted.
+        """
+        claims = {"course_id": exam.course_id, "content_id": exam.content_id}
+        expiration_window = 60
+        exam_attempt = cls.get_current_exam_attempt(user, exam)
+        cookie_name = "exam_access_ticket"
+
+        data = {}
+        grant_access = True
```
varshamenon4, line 341: Question: for readability, do you think it is better to set grant_access to false, then change to true when access should be granted? Kind of a nit but just wanted to get another take.

varshamenon4 (commented):
Ready for final review! Added some questions. Also curious if there is a good way to test this manually/locally.

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +274,116 @@ class ProctoringProvidersView(ListAPIView):
...
+        Given a user and exam, get the user's latest exam attempt, if exists.
+        """
+        try:
+            exam_attempt = ExamAttempt.objects.filter(user=user, exam=exam).latest('created')
+        except ObjectDoesNotExist:
+            exam_attempt = None
+        return exam_attempt
+
+    @classmethod
+    def set_expiration_window(cls, exam_attempt, default_exp_seconds):
```
Zacharis278, line 310: Looks like this is more of a `get_expiration_window` or a `calculate_expiration_window` since it just returns a value

edx_exams/apps/api/v1/views.py
Zacharis278: this makes sense to me. I'd suggest a 404 here if the exam id is not found however.

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +274,116 @@ class ProctoringProvidersView(ListAPIView):
...
+        claims = {"course_id": exam.course_id, "content_id": exam.content_id}
+        expiration_window = 60
+        exam_attempt = cls.get_current_exam_attempt(user, exam)
+        cookie_name = "exam_access_ticket"
+
+        data = {}
+        grant_access = True
+        response_status = status.HTTP_200_OK
+
+        # If exam attempt exists for user, then grant exam access
```
Zacharis278, line 344: Having a hard time following this block. If the exam attempt doesn't exist at all maybe we could just short circuit and return a 404 first? That might make structuring the rest of this easier to read.

What about created, ready, submitted, etc attempts? Seems to me they just fall through all of these cases and get access by default.

edx_exams/apps/api/v1/views.py
Zacharis278: yes, but we should probably introduce a constant/enum for attempt statuses. Example https://github.com/openedx/edx-proctoring/blob/0ae746c253db5910cd6f91abf60e19771102130c/edx_proctoring/statuses.py#L8

edx_exams/settings/test.py
```diff
@@ -23,3 +23,47 @@
 )
 
 LMS_ROOT_URL = 'http://test.lms:18000'
+
+TOKEN_SIGNING = {
```
Zacharis278, line 27: we'll need to have this setting setup to work in other environments too. Typically what we'd do is put this in base.py and than override it in the local/test/production files. We can take a look at this after some of the other comments above tho since it's just config.

As an example https://github.com/edx/edx-exams/blob/2ca064b98a5c77f40b9583d1ab7147ad250bb499/edx_exams/settings/base.py#L218

Zacharis278 (changes requested):

edx_exams/apps/api/v1/views.py
varshamenon4: That makes sense, will do.

edx_exams/apps/api/v1/views.py
varshamenon4: Good point.

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +274,116 @@ class ProctoringProvidersView(ListAPIView):
...
+        * exam_id: This is the id of the exam that the user is requesting access to
+    **Exceptions**
+        * HTTP_403_FORBIDDEN
+    """
+
+    authentication_classes = (JwtAuthentication,)
+    permission_classes = (StaffUserOrReadOnlyPermissions,)
+
+    @classmethod
+    def get_current_exam_attempt(cls, user, exam):
```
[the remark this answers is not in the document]
varshamenon4: Yeah, so I had initially tried that and ran into some bugginess while testing. I'll try that again and see if I'm able to get working.

edx_exams/apps/api/v1/views.py
varshamenon4: Ah maybe I didn't understand the logic correctly here then. I thought that if the exam attempt doesn't exist, but the exam is past due, that the person should still get access?

Oh, yeah. I didn't really realize those were potential cases. If they're created, ready, submitted, what's the expected behavior? Should we only give access if started, or are some of these other cases also ones where access should be granted?

Zacharis278: oh yeah thinking about this more my original comment doesn't even cover the whole scope of this. We also differentiate behavior between timed and proctored exams. See https://2u-internal.atlassian.net/wiki/spaces/PT/pages/15440919/Exam+Visibility+Behavior

You can also take a look at the code that deals with this today https://github.com/openedx/edx-proctoring/blob/0ae746c253db5910cd6f91abf60e19771102130c/edx_proctoring/api.py#L2958. There's a bunch of other complexities beyond 'does the learner have access' in that code though.

Honestly, I'm happy to make the decision to just simplify this and get rid of the proctored vs. timed differentiation. We can always add that back later if it creates problems but there's been a desire by the support team to change this anyway.

My suggestion:
- Before the due date exams should only ever return in access token if the attempt is in the 'started' state.
- After the due date exams would allow an access token if the attempt is 'submitted' (if timed) or 'verified' (if proctored) after due. This only applies if exam.hide_after_due is False, otherwise no access token

varshamenon4: This is a more basic question about exams, but do we need to consider the case where it is a practice exam and if so what would the access look like? Would we always grant access? Or is this case not relevant here? If not, how are we certain that we won't get practice exams for for the exam access endpoint?

The main reason I ask is because I noticed in the Exam Visibility Behavior doc that all special exams (proctored and timed) have due dates. So I am wondering if I need to actually check if the exam has a due date first or if I can just check what the due date is relative to now.

Zacharis278: I don't think a practice exam would change the way this works since it should behave the same as it's proctored or timed counterpart, it's just not graded. As far as due date, we should check against that because it is possible an exam may not actually have a due date. I think the doc just doesn't call it out but those exams are just perpetually 'before the due date'.

edx_exams/settings/test.py
varshamenon4: @Zacharis278 I see, good point. So I had grabbed the TOKEN_SIGNING from the token-utils repo: https://github.com/edx/token-utils/blob/main/test_settings.py#L64

But these are specifically only used for the test settings. For base.py, should I set these to none and then have them overridden with something else for local/production? Specifically wondering about JWT_PUBLIC_SIGNING_JWK_SET and JWT_PRIVATE_SIGNING_JWK as the other stuff looks like it can be left as is.

edx_exams/apps/access/views.py
varshamenon4: @Zacharis278 can this also go in base.py like the jwt token signing below?

Zacharis278: @varshamenon4 yup I think we need a setting for both ACCESS_TOKEN_COOKIE_NAME and ACCESS_TOKEN_COOKIE_DOMAIN. The name could be set there and the domain should probably be None.

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +275,104 @@ class ProctoringProvidersView(ListAPIView):
...
+    def get_response(cls, exam, user):
+        """
+        Get response, with exam access ticket if access granted.
+
+        403 error if access is not granted.
+        """
+        claims = {"course_id": exam.course_id, "content_id": exam.content_id}
+        expiration_window = 60
+        exam_attempt = ExamAttempt.get_current_exam_attempt(user, exam)
+        cookie_name = "exam_access_ticket"
```
Zacharis278, line 328: see my other comment. Let's use a setting for this

edx_exams/settings/test.py
Zacharis278: ok think I've got my head around this, base.py should include the token settings we always need so something like this

```
TOKEN_SIGNING = {
    'JWT_ISSUER': 'http://127.0.0.1:8740/',
    'JWT_SIGNING_ALGORITHM': 'RS512',
    'JWT_SUPPORTED_VERSION': '1.2.0',
    'JWT_PRIVATE_SIGNING_JWK': 'replace-me',
}
```

To deal with our dev environments we'll just update JWT_PRIVATE_SIGNING_JWK to have a value in local.py. (see that local.py imports from base.py)

Notice I didn't include the JWT_PUBLIC_SIGNING_JWK_SET. edx-exams doesn't need to validate this JWT so it doesn't need that setting however we should include it in test.py. So in test we'll need to do an update on both JWT_PRIVATE_SIGNING_JWK and JWT_PUBLIC_SIGNING_JWK_SET.

Finally for prod/stage we need to have TOKEN_SIGNING included in https://github.com/edx/edx-exams/blob/b02c4f373471e0ad175a56ddb4c96c121008513e/edx_exams/settings/production.py#L17 so we can add our settings to edx-internal.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -513,3 +514,332 @@ def test_proctoring_providers_list_empty(self):
...
+            attempt_number=1,
+            status='verified',
+            start_time=start_time,
+            allowed_time_limit_mins=allowed_time_limit_mins
+        )
+
+        response = self.get_exam_access(self.user, past_due_url)
+        self.assertEqual(403, response.status_code)
+
+    def test_access_not_granted_not_started_exam_attempt(self):
```
Zacharis278, line 759: no need to refactor unless you want to but most of these tests are performing the same logic but with different inputs for attempt status, past due, and expected response. You can use [ddt](https://ddt.readthedocs.io/en/latest/) to test a bunch of these all at once.

Take a look at usage in edx-proctoring as an example
https://github.com/openedx/edx-proctoring/blob/0ae746c253db5910cd6f91abf60e19771102130c/edx_proctoring/tests/test_api.py#L1876

edx_exams/apps/api/utils.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+import logging
+from datetime import timedelta
+
+from django.utils import timezone
+
+log = logging.getLogger(__name__)
+
+
+def get_exam_attempt_time_remaining(start_time, time_limit_mins, now=None):
```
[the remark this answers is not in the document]
varshamenon4: That's a good point @Zacharis278. I'm trying to understand what would happen if either/both the start time or time_limit are null. Here are my thoughts:

- if the time limit is null, then we could consider the time limit infinite and use infinity value for that. But I worry that this only works for the current case (since it will use the shorter of the time remaining vs the expiration window)
- if the start time is null, maybe the time limit would be the same as the time limit mins or maybe the time remaining would be some large value again?
- alternatively, and maybe this is the simplest solution, we could have the function return an exception if either are null.

For now I'm going to implement assuming the last one but lemme know if you think otherwise.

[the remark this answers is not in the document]
varshamenon4: Actually what I think I'll do is raise an error if either are null and keep the expiration window the same if the error is thrown.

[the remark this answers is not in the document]
Zacharis278: for most of these types of questions my default is always to just look at what we do today and see if there's any justification for changing the behavior. In this case the remaining time is zero if they have not started or an end has not yet been set. Which, logically makes sense to me.

https://github.com/openedx/edx-proctoring/blob/0ae746c253db5910cd6f91abf60e19771102130c/edx_proctoring/utils.py#L48

edx_exams/apps/api/v1/views.py
```diff
@@ -5,17 +5,22 @@
...
 
 from django.core.exceptions import ObjectDoesNotExist
+from django.utils import timezone
 from edx_api_doc_tools import path_parameter, schema
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from rest_framework import status
 from rest_framework.generics import ListAPIView
 from rest_framework.response import Response
 from rest_framework.views import APIView
+from token_utils.sign import create_jwt
```
Zacharis278, line 15: just noticed that this isn't using the public API for token utils. We should be using the api here https://github.com/edx/token-utils/blob/main/token_utils/api.py#L8

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -513,3 +514,279 @@ def test_proctoring_providers_list_empty(self):
...
+        """
+        response = self.get_exam_access(self.user, self.url)
+        self.assertEqual(403, response.status_code)
+
+    @ddt.data(
+        ('started', 200),
+        ('created', 403),
+    )
+    @ddt.unpack
+    def test_access_granted_started_exam_attempt(self, attempt_status, response_status):
```
Zacharis278, line 601: This test and `test_access_granted_verified_attempt_status` appear to be exactly the same other than the exam object. Any reason not to just parameterize exam_past_due as an input and collapse these into a single function?

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +276,106 @@ class ProctoringProvidersView(ListAPIView):
...
+
+        data = {"detail": "Exam access ticket not granted"}
+        grant_access = False
+        response_status = status.HTTP_403_FORBIDDEN
+
+        # If exam attempt exists for user, then grant exam access
+        if exam_attempt is not None:
+            # If no due date, grant access if attempt started
+            if exam.due_date is None:
+                if exam_attempt.status == ExamAttemptStatus.started:
```
Zacharis278, line 337: if there's no due date don't we still need to get_expiration_window since that would take into account the time remaining on the attempt which might be less than the default expiration_window

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -513,3 +514,279 @@ def test_proctoring_providers_list_empty(self):
...
+            exam=self.exam,
+            attempt_number=1,
+            status='started',
+            start_time=start_time,
+            allowed_time_limit_mins=allowed_time_limit_mins
+        )
+
+        response = self.get_exam_access(self.user, self.url)
+        self.assertEqual(200, response.status_code)
+        self.assertIsInstance(response.cookies["exam_access_ticket"].value, str)
```
Zacharis278, line 783: noticed we might be missing some validation here. We should be checking if the expiration on this token actually matches the expected value. I think this is also true for the lms_user_id and the content_id, I don't see any tests making sure those are correct.

edx_exams/settings/test.py
```diff
@@ -23,3 +23,44 @@
...
+        RrqDiFyzrJYivodqTQ",
+        "q": "3T3DEtBUka7hLGdIsDlC96Uadx_q_E4Vb1cxx_4Ss_wGp1Loz3N3ZngGyInsKlmbBgLo1Ykd6T9TRvRNEWEtFSOcm2INIBoVoXk7W5Ru\
+        Pa8Cgq2tjQj9ziGQ08JMejrPlj3Q1wmALJr5VTfvSYBu0WkljhKNCy1KB6fCby0C9WE",
+        "p": "vUqzWPZnDG4IXyo-k5F0bHV0BNL_pVhQoLW7eyFHnw74IOEfSbdsMspNcPSFIrtgPsn7981qv3lN_staZ6JflKfHayjB_lvltHyZxfl0\
+        dvruShZOx1N6ykEo7YrAskC_qxUyrIvqmJ64zPW3jkuOYrFs7Ykj3zFx3Zq1H5568G0",
+        "kid": "token-test-sign", "kty": "RSA"
+    }''',
+    'JWT_PUBLIC_SIGNING_JWK_SET': '''{
+        "keys": [
+            {
```
Zacharis278, line 46: we definitely don't need this wrong key here but maybe keep the second one if we want to use the unpack function in tests

Zacharis278 (commented):
awesome! I think this is super close, we're just missing a small amount of test coverage validating the token content/expiration.

edx_exams/apps/api/v1/tests/test_views.py
varshamenon4: Good point! The reason I didn't is because adding self.exam as a parameter was returning unresolved reference 'self'/NameError: name 'self' is not defined. I did check in the original proctoring example you'd sent, and didn't see any examples of parameterizing objects instantiated in setup. I could create two new exam objects as parameters but I figured this would be redundant which is why I kept as two separate tests. Let me know if there is another way to approach.

edx_exams/apps/api/v1/views.py
varshamenon4: Ah good point. Just fixed.

edx_exams/apps/api/v1/tests/test_views.py
varshamenon4: Ah yes you're right, so this would require unpacking the token in the cookie and getting these values right?

edx_exams/apps/api/v1/tests/test_views.py
Zacharis278: Yeah, generally you want to keep these parameters to primitives (strings, numbers, booleans) because they actually become part of the test name. (the first test run here is actually `test_access_granted_started_exam_attempt_started_200`) so the way to do this is just pass a boolean for exam_past_due and setup based on that.

eg:
```
def test_access_granted_started_exam_attempt(self, attempt_status, response_status, exam_past_due):
    exam = self.past_due_exam if exam_past_due else self.exam
    etc...
```

edx_exams/settings/test.py
varshamenon4: Ah good point. That was probably just for token-utils testing. Removed that but keeping the other one for the unpack function for testing expiration window, lms user id, and content id.

edx_exams/apps/api/utils.py
```diff
@@ -0,0 +1,34 @@
...
+        return 0
+
+    now = now or timezone.now()
+
+    # Calculate end of attempt time
+    end_time = exam_attempt.start_time + timedelta(minutes=exam_attempt.allowed_time_limit_mins)
+
+    # Compare end of attempt time to now
+    # If end time has passed, set to zero
+    if end_time > now:
```
varshamenon4, line 28: I had the logic on this wrong (`end_time < now` before), so I fixed it. Could you take a look to make sure that looks right. Also updated test utils accordingly.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -513,3 +515,265 @@ def test_proctoring_providers_list_empty(self):
...
+        (True, 'verified', 200),
+        (True, 'created', 403),
+    )
+    @ddt.unpack
+    def test_access_granted_started_exam_attempt(self, exam_past_due, attempt_status, response_status):
+        """
+        Verify the endpoint grants/doesn't grant access for an exam
+        based on the status of the exam attempt.
+        """
+        exam = self.past_due_exam if exam_past_due else self.exam
```
varshamenon4, line 614: Added this per your suggestion.

varshamenon4 (commented):
Ready for final final final review. ;P
Added a couple of comments.

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +276,103 @@ class ProctoringProvidersView(ListAPIView):
...
+        elif exam.due_date is not None and timezone.now() >= exam.due_date:
+            data, grant_access, response_status = {}, True, status.HTTP_200_OK
+
+        response = Response(status=response_status,
+                            data=data)
+
+        if grant_access:
+            log.info("Creating exam access ticket")
+            access_ticket = sign_token_for(user.lms_user_id, expiration_window, claims)
+            response.set_cookie(ACCESS_TOKEN_COOKIE_NAME, access_ticket)
```
Zacharis278, line 359: I missed this on the first few go arounds but this needs to be set on .edx.org and not exams.edx.org
```suggestion
            response.set_cookie(ACCESS_TOKEN_COOKIE_NAME, access_ticket, domain=settings.ACCESS_TOKEN_COOKIE_DOMAIN)
```

edx_exams/settings/local.py
```diff
@@ -89,6 +89,26 @@
...
+        3A1diLRCrKHOvnpvgQ",
+        "n": "o5cn3ljSRi6FaDEKTn0PS-oL9EFyv1pI7dRgffQLD1qf5D6sprmYfWWokSsrWig8u2y0HChSygR6Jn5KXBqQn6FpM0dDJLnWQDRXHLl3\
+        Ey1iPYgDSmOIsIGrV9ZyNCQwk03wAgWbfdBTig3QSDYD-sTNOs3pc4UD_PqAvU2nz_1SS2ZiOwOn5F6gulE1L0iE3KEUEvOIagfHNVhz0oxa_V\
+        RZILkzV-zr6R_TW1m97h4H8jXl_VJyQGyhMGGypuDrQ9_vaY_RLEulLCyY0INglHWQ7pckxBtI5q55-Vio2wgewe2_qYcGsnBGaDNbySAsvYcW\
+        RrqDiFyzrJYivodqTQ",
+        "q": "3T3DEtBUka7hLGdIsDlC96Uadx_q_E4Vb1cxx_4Ss_wGp1Loz3N3ZngGyInsKlmbBgLo1Ykd6T9TRvRNEWEtFSOcm2INIBoVoXk7W5Ru\
+        Pa8Cgq2tjQj9ziGQ08JMejrPlj3Q1wmALJr5VTfvSYBu0WkljhKNCy1KB6fCby0C9WE",
+        "p": "vUqzWPZnDG4IXyo-k5F0bHV0BNL_pVhQoLW7eyFHnw74IOEfSbdsMspNcPSFIrtgPsn7981qv3lN_staZ6JflKfHayjB_lvltHyZxfl0\
+        dvruShZOx1N6ykEo7YrAskC_qxUyrIvqmJ64zPW3jkuOYrFs7Ykj3zFx3Zq1H5568G0",
+        "kid": "token-test-sign", "kty": "RSA"
```
Zacharis278, line 108: can we come up with a id for this that describes what it's used for?
```suggestion
        "kid": "localdev_exam_token_key", "kty": "RSA"
```

Zacharis278 (approved):
two really minor things otherwise LGTM. We'll also need to spin up a PR into https://github.com/edx/edx-internal/tree/master/argocd/applications/edx-exams for any settings changes but I think this can merge first w/o causing problems.

PR #67, opened 2023-01-13
fix: LTI link url should not be 'internal'

Zacharis278:
Bug found while testing LTI launch on stage. The host url for this launch would end up being edx-exams-internal.edx.org which would cause the session cookie to get set on the internal domain. We want to launch to edx-exams.edx.org so any redirect from the tool (which would use edx-exams.edx.org) includes the session cookie.


varshamenon4 (approved):
lgtm!

PR #69, opened 2023-01-25
feat: make django admin not terrible

Zacharis278:
**JIRA:** https://2u-internal.atlassian.net/browse/MST-1731

- Fixes the blank labels, empty dropdowns, and infinite horizontal scrolling
- Added readonly to some fields that really shouldn't be updated in admin

### DIFF
--- a/edx_exams/apps/core/admin.py
+++ b/edx_exams/apps/core/admin.py
@@ -30,17 +30,20 @@ class ProctoringProviderAdmin(admin.ModelAdmin):
 
 class ExamAdmin(admin.ModelAdmin):
     """ Admin configuration for the Exam model """
-    list_display = ('resource_id', 'course_id', 'provider', 'content_id', 'exam_name',
-                    'exam_type', 'time_limit_mins', 'due_date', 'hide_after_due', 'is_active')
+    list_display = ('course_id', 'provider', 'exam_name', 'exam_type', 'due_date', 'is_active')
+    readonly_fields = ('resource_id', 'course_id')
+    list_filter = ('is_active',)
     search_fields = ('resource_id', 'course_id', 'provider__name', 'content_id', 'exam_name',
                      'exam_type', 'time_limit_mins', 'due_date', 'hide_after_due', 'is_active')
-    ordering = ('course_id', 'exam_name',)
+    ordering = ('-is_active', 'course_id', 'exam_name',)
 
 
 class ExamAttemptAdmin(admin.ModelAdmin):
     """ Admin configuration for the Exam Attempt model """
     list_display = ('user', 'exam', 'attempt_number', 'status', 'start_time',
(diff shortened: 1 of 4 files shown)


edx_exams/apps/core/models.py
```diff
@@ -44,9 +45,6 @@ class Meta:
     def get_full_name(self):
         return self.full_name or super().get_full_name()
 
-    def __str__(self):
```
Zacharis278, line 47: when printing user we're usually going to prefer the default behavior (username) for admin pages, logging, etc so I've removed this.

edx_exams/apps/core/admin.py
```diff
@@ -30,17 +30,20 @@ class ProctoringProviderAdmin(admin.ModelAdmin):
 
 class ExamAdmin(admin.ModelAdmin):
     """ Admin configuration for the Exam model """
-    list_display = ('resource_id', 'course_id', 'provider', 'content_id', 'exam_name',
-                    'exam_type', 'time_limit_mins', 'due_date', 'hide_after_due', 'is_active')
+    list_display = ('course_id', 'provider', 'exam_name', 'exam_type', 'due_date', 'is_active')
```
alangsto, line 33: Just curious why `time_limit_mins` and `hide_after_due` aren't included here. Are those fields not included in the current admin pages for edx-proctoring? Or is the information not really relevant?

edx_exams/apps/core/admin.py
```diff
@@ -30,17 +30,20 @@ class ProctoringProviderAdmin(admin.ModelAdmin):
...
-    ordering = ('course_id', 'exam_name',)
+    ordering = ('-is_active', 'course_id', 'exam_name',)
 
 
 class ExamAttemptAdmin(admin.ModelAdmin):
     """ Admin configuration for the Exam Attempt model """
     list_display = ('user', 'exam', 'attempt_number', 'status', 'start_time',
                     'allowed_time_limit_mins')
+    list_filter = ('status',)
+    readonly_fields = ('user', 'exam')
```
alangsto, line 46: should `attempt_number` also be read only?

alangsto (approved):
Just a few questions, otherwise it looks good 👍

edx_exams/apps/core/admin.py
Zacharis278: Yeah we don't include these today. This is just the list view with a table of every exam, you can see the attributes when you click into the page for that individual record. I dropped most fields off here since more than a few columns won't fit before scrolling off the the right.

edx_exams/apps/core/admin.py
Zacharis278: this is the id according the the LTI tool so I figured I'd leave this editable for testing's sake you can't really break anything by changing it right now. We might need to lock this down at some point tho.

PR #71, opened 2023-01-30
feat: automatically use jwt auth

Zacharis278:
This may be a helpful fix as we build out browser endpoints in M3. Right now in order to test our JWT endpoints without a frontend it requires a browser extension like requestly to apply the USE_JWT_COOKIE_HEADER (example - https://app.requestly.io/rules/#sharedList/1671117948648-apply-edx-exams-jwt). This would make it so that header is implied if a JWT cookie is sent by the browser.

This is a bit jankey... but may be removed once we've completed the UI work.

### DIFF
--- a/edx_exams/apps/core/middleware.py
+++ b/edx_exams/apps/core/middleware.py
@@ -1,0 +1,23 @@
+"""
+Middleware that checks if in incoming request has a browser jwt cookie
+and enables JWT auth for that request.
+
+This is a temporary workaround that allows easier testing of browser endpoints in
+absence of a frontend UI. Normally a frontend application must explicity request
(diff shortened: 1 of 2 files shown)


ashultz0 (approved):
I think we should just have this everywhere for every service because we finished migrating to the new cookies a long time ago.

alangsto (approved):

Zacharis278:
@ashultz0 I'll make a note to bring this up. There's some weirdness where the initial JWT validation don't know if the endpoint we're hitting actually expects JWT auth and may reject if we're on a page that doesn't refresh it. (basically any page that isn't an MFE yet)

PR #72, opened 2023-02-01
feat: add endpoint to allow attempt updates

alangsto:
**JIRA:** https://2u-internal.atlassian.net/browse/MST-1641

**Description:** This PR adds an endpoint to support PUT requests meant to update an attempt's status. While the edx-proctoring equivalent adds in many behaviors to this endpoint, support for those behaviors in edx-exams is limited right now. This means that the ability to decline an attempt, retry an attempt, and reset an attempt are not supported at this time. 

Additionally, updating an attempt's status triggers downstream effects for name affirmation, grading, credentials, and completion, but those are also scoped out of this ticket.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**


edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +273,79 @@ class ProctoringProvidersView(ListAPIView):
...
+            'stop': ExamAttemptStatus.ready_to_submit,
+            'start': ExamAttemptStatus.started,
+            'submit': ExamAttemptStatus.submitted,
+            'click_download_software': ExamAttemptStatus.download_software_clicked,
+            'error': ExamAttemptStatus.error,
+        }
+
+        to_status = action_mapping.get(action)
+        if to_status:
+            allow_transition, error_msg = allow_status_transition(attempt_id, to_status)
```
alangsto, line 341: Not sure if it's better to check if a transition is allowed in the view or in `update_attempt_status`.

edx_exams/apps/api/v1/views.py
```diff
@@ -271,3 +273,79 @@ class ProctoringProvidersView(ListAPIView):
...
+            if allow_transition:
+                attempt_id = update_attempt_status(attempt_id, to_status)
+            else:
+                return Response(
+                    status=status.HTTP_400_BAD_REQUEST,
+                    data={'detail': error_msg}
+                )
+
+        data = {"exam_attempt_id": attempt_id}
```
alangsto, line 350: edx-proctoring has an odd behavior where a 200 response is returned even when an action isn't recognized. I've chosen that same behavior here for consistency, but am wondering if that's something we want to change (i.e. return 400 instead)

edx_exams/apps/core/api.py
```diff
@@ -0,0 +1,80 @@
...
+
+import pytz
+
+from edx_exams.apps.core.models import ExamAttempt
+from edx_exams.apps.core.statuses import ExamAttemptStatus
+
+log = logging.getLogger(__name__)
+
+
+def update_attempt_status(attempt_id, to_status):
```
alangsto, line 15: this is an extremely paired down version of what exists in edx-proctoring. I would like to keep it this way if possible. Any additional logic that needs to be added for grading, credentials, completion, and name verification should be added outside of this function so we can limit the scope of this to only updating the attempt status. This function should make the calls to trigger those behaviors, rather than define those behaviors.

edx_exams/apps/core/api.py
```diff
@@ -0,0 +1,80 @@
...
+            'user_id': user_id,
+            'from_status': from_status,
+            'to_status': to_status,
+        }
+    )
+
+    attempt_obj.status = to_status
+
+    if to_status == ExamAttemptStatus.started:
+        attempt_obj.start_time = datetime.now(pytz.UTC)
```
alangsto, line 41: The allowed time is also calculated here, working on that now

edx_exams/apps/api/v1/views.py
Zacharis278: I'd think there is going to be other places we try to `update_attempt_status` so my gut is it would be cleaner to put this there where the check always happens and is centralized. Maybe we define a custom exception to throw if the update is invalid and just catch it here?

edx_exams/apps/api/v1/views.py
Zacharis278: I don't see any reason why we'd receive junk actions with this new UI so I'd say just return the 400.

edx_exams/apps/core/api.py
Zacharis278: 1000% agreed. Maybe we even use signals/celery to further decouple these behaviors from the UI request thread? Or event bus if that's a viable option. Later problems either way

Zacharis278 (commented):
I really like using the core/api file to hold the main functionality of interacting with exam attempts. We could actually take this further if we wanted and have the api entirely wrap the models so the rest views don't import them directly. This is what the program_enrollments / grades apps in the LMS do. It would require some additional refactoring on some existing endpoint tho.

edx_exams/apps/core/api.py
Zacharis278: related, I think we should now move this into your core/api file https://github.com/edx/edx-exams/blob/main/edx_exams/apps/api/utils.py#L12

Zacharis278 (commented):
Generally looking good to me.

One thing I'm noticing on the tests, we're getting a lot of our test coverage from the rest view calling down into the api interface. It might be cleaner down the line to test those things separately if the python api gets more stuff added into it. We follow the current pattern you have all over the place so not a problem, just something to think about in future features.

edx_exams/apps/api/v1/views.py
```diff
@@ -376,3 +379,76 @@ def get(self, request, exam_id):
...
+        attempt = ExamAttempt.get_attempt_by_id(attempt_id)
+        action = request.data.get('action')
+
+        if not attempt:
+            return Response(
+                status=status.HTTP_400_BAD_REQUEST,
+                data={'detail': f'Attempt with attempt_id={attempt_id} does not exit.'}
+            )
+
+        # user should only be able to update their own attempt if non staff
```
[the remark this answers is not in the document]
alangsto: yes, thanks for catching that comment

alangsto:
@Zacharis278 I think it's better to start now, so I will move some of the existing test logic into the api tests. 

I'm also wondering how much refactoring should happen in this PR to ensure that objects are only handled in api.py. I can clean it up for the view I just wrote, but maybe it's worth creating a separate ticket to do the other views? Should any of this be documented in an ADR as well?

Zacharis278:
@alangsto maybe we just document the intent in the docstring for api.py? We already kinda have an ADR for this to reference, technically it's specific to edx-platform but the point is the same https://github.com/openedx/edx-platform/blob/40467d86049132949acaee23f38718a1076ab643/docs/decisions/0002-inter-app-apis.rst

No strong opinions on when we refactor, I'm totally fine with chunking off relevant bits as we go.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -745,28 +745,31 @@ def test_access_not_granted_no_status_exam_attempt(self):
         self.assertEqual(403, response.status_code)
 
     @ddt.data(
-        (-timedelta(minutes=25), 403),  # exam attempt with zero time remaining
-        (timedelta(minutes=5), 403),  # exam attempt that is past due date
-        (-timedelta(seconds=2140), 200),  # exam attempt time remaining less than due date
+        (timedelta(minutes=35), timedelta(minutes=0), 403),  # exam attempt with zero time remaining
+        (timedelta(minutes=10), timedelta(minutes=10), 403),  # exam attempt that is past due date
+        (timedelta(minutes=20), timedelta(minutes=0), 200),  # exam attempt time remaining less than due date
```
varshamenon4, line 750: Nice, this is a clever approach!

varshamenon4 (approved):
The updated test logic looks great! And the fix to the time remaining.

alangsto:
@Zacharis278 added additional testing for the API functions, and updated logic so that the view does not handle the attempt object at all. Any API functions used by the view take in an attempt_id, rather than an object, while helper functions defined within `api.py` are allowed to have objects passed in. 

I also fixed the logic error for the access ticket code, and updated some of the tests that were failing as a result of that fix.

edx_exams/apps/core/tests/test_api.py
```diff
@@ -0,0 +1,269 @@
...
+        """
+        update_attempt_status(self.exam_attempt.id, ExamAttemptStatus.started)
+
+        with self.assertRaises(ExamIllegalStatusTransition) as exc:
+            # update again to started
+            update_attempt_status(self.exam_attempt.id, ExamAttemptStatus.started)
+            self.assertIn('Cannot start exam attempt', exc)
+
+    @ddt.data(
+        *product(ExamAttemptStatus.completed_statuses, ExamAttemptStatus.incomplete_statuses)
```
Zacharis278, line 213: neat, I didn't know you could do this

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -777,3 +782,163 @@ def test_access_granted_past_due_exam_no_attempt(self):
         response = self.get_exam_access(self.user, self.past_due_url)
         self.assertEqual(200, response.status_code)
         self.assert_valid_exam_access_ticket(response, self.user, self.past_due_exam)
+
+
+@ddt.ddt
+class ExamAttemptViewTest(ExamsAPITestCase):
```
Zacharis278, line 788: nit: since we're already testing the api functions separately this could get pared down a lot with some mocks if you wanted.

Zacharis278 (approved):
looking good to me 👍

edx_exams/apps/api/v1/tests/test_views.py
alangsto: ohhh interesting 🤔 I'll see what I can do

PR #73, opened 2023-02-10
feat: add endpoint to create exam attempt

alangsto:
**JIRA:** [MST-1640](https://2u-internal.atlassian.net/browse/MST-1640)

**Description:** This PR adds support for an endpoint that allows exam attempt creation. The endpoint expects the following data 
`{ 
  exam_id: <id>, 
  start_clock: <Boolean>, 
}`

If the endpoint is called for a non-practice exam whose due date has passed, an exception is thrown. To mimic edx-proctoring behavior, attempts can be created for practice exams whose due date has passed.

**Merge checklist:**
- [ ] All reviewers approved
- [x] CI build is green
- [ ] Changelog record added
- [x] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -15,7 +15,7 @@ from token_utils.api import unpack_token_for
 
 from edx_exams.apps.api.test_utils import ExamsAPITestCase
 from edx_exams.apps.api.test_utils.factories import UserFactory
-from edx_exams.apps.core.exceptions import ExamIllegalStatusTransition
+from edx_exams.apps.core.exceptions import ExamAttemptOnPastDueExam, ExamIllegalStatusTransition
 from edx_exams.apps.core.models import CourseExamConfiguration, Exam, ExamAttempt, ProctoringProvider
 from edx_exams.apps.core.statuses import ExamAttemptStatus
 
@@ -830,7 +830,17 @@ class ExamAttemptViewTest(ExamsAPITestCase):
 
         return self.client.put(url, data, **headers, content_type="application/json")
 
-    def test_user_update_permissions(self):
+    def post_api(self, user, data):
+        """
+        Helper function to make post request to the API
+        """
+        data = json.dumps(data)
+        headers = self.build_jwt_headers(user)
+        url = reverse('api:v1:exams-attempt')
+
+        return self.client.post(url, data, **headers, content_type="application/json")
+
+    def test_put_user_update_permissions(self):
         """
         Test that non-staff users cannot update the attempt of another user
         """
@@ -849,7 +859,7 @@ class ExamAttemptViewTest(ExamsAPITestCase):
         response = self.put_api(self.non_staff_user, attempt.id, {'action': 'start'})
         self.assertEqual(response.status_code, 403)
 
-    def test_attempt_does_not_exist(self):
+    def test_put_attempt_does_not_exist(self):
         """
         Test that updating an attempt that does not exist fails
         """
@@ -865,7 +875,7 @@ class ExamAttemptViewTest(ExamsAPITestCase):
     )
     @ddt.unpack
     @patch('edx_exams.apps.api.v1.views.update_attempt_status')
-    def test_update_exam_attempt(self, action, expected_status, mock_update_attempt_status):
+    def test_put_update_exam_attempt(self, action, expected_status, mock_update_attempt_status):
         """
         Test that an exam can be updated
         """
@@ -886,7 +896,7 @@ class ExamAttemptViewTest(ExamsAPITestCase):
         mock_update_attempt_status.assert_called_once_with(attempt.id, expected_status)
 
     @patch('edx_exams.apps.api.v1.views.update_attempt_status')
-    def test_exception_raised(self, mock_update_attempt_status):
+    def test_put_exception_raised(self, mock_update_attempt_status):
         """
         Test that if an exception is raised, endpoint returns 400 with error message
         """
@@ -906,7 +916,7 @@ class ExamAttemptViewTest(ExamsAPITestCase):
         self.assertEqual(response.status_code, 400)
         self.assertEqual(error_msg, response.data['detail'])
 
-    def test_invalid_action(self):
+    def test_put_invalid_action(self):
         """
         Test that an unrecognized action fails
         """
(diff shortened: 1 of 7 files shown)


edx_exams/apps/core/api.py
```diff
@@ -142,3 +144,58 @@ def get_exam_attempt_time_remaining(exam_attempt, now=None):
...
+            f"exam_id={exam_id} in course_id={exam_obj.course_id}. Do not register an exam attempt!"
+        )
+        raise ExamAttemptOnPastDueExam(err_msg)
+
+    # create exam attempt
+    attempt = ExamAttempt.objects.create(
+        exam_id=exam_id,
+        user_id=user_id,
+        status=ExamAttemptStatus.created,
+        attempt_number=1,
```
alangsto, line 188: this needs a non-null value, but I'm not quite sure what the value should be. This is an int field, and it's needed by the LTI code, but I'm not sure if it should be auto-incrementing or if we should create a random int value.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -830,7 +830,17 @@ def put_api(self, user, attempt_id, data):
...
+        """
+        Helper function to make post request to the API
+        """
+        data = json.dumps(data)
+        headers = self.build_jwt_headers(user)
+        url = reverse('api:v1:exams-attempt')
+
+        return self.client.post(url, data, **headers, content_type="application/json")
+
+    def test_put_user_update_permissions(self):
```
alangsto, line 843: renamed these to add `put` because I was adding tests for the same endpoint

edx_exams/apps/core/api.py
MichaelRoytman: I'll quote the relevant [portion](http://www.imsglobal.org/spec/proctoring/v1p0#h.m0wu6zl84ted) of the proctoring specification below.

> The https://purl.imsglobal.org/spec/lti-ap/claim/attempt_number claim MUST specify the candidate's attempt number.  Used in conjunction with the resource link claim this information can be used by the Proctoring Tool to determine if the candidate is starting a new attempt at the assessment or is returning to resume a previous attempt.  The value SHOULD be simple integer with the first attempt having attempt number 1.
> 
> This specification imposes no limits on the number of times a Candidate can initiate the proctoring process for a given attempt at an assessment.  A Proctoring Tool MAY impose a limit of one successful launch per assessment (resource_link) per user.  Multiple attempts at the same assessment SHOULD NOT be assumed without prior arrangement with the Proctoring Tool.
>
> Example 4.2.3.6.1: the Candidate initiates the exam launch workflow and is transferred to the Proctoring Tool.  Once connected with the [proctor](http://www.imsglobal.org/spec/proctoring/v1p0#dfn-proctor) they are asked to present their government issued ID document but are unable to provide anything suitable.  They return the following day and initiate the exam launch workflow a second time.  The Assessment Platform sends the sameattempt number in both Start Proctoring messages.
> 
> Example 4.2.3.6.2: the Candidate initiates the assessment launch workflow for their first attempt at their certification exam.  The Assessment Platform sets the attempt number to 1 in the Start Proctoring message.  The [proctor](http://www.imsglobal.org/spec/proctoring/v1p0#dfn-proctor) is satisfied that the candidate's environment is secure and the candidate is allowed to start the assessment.  The candidate fails the exam and returns one month later.  When they initiate the launch workflow again the Assessment Platform sends attempt number 2.

Based on this, I think the value should be an auto-incrementing integer starting at 1. The specification doesn't seem to outright require that it be auto-incrementing, but it does suggest it.

Does that help?

alangsto: I think it does, but are they saying that the attempt_number is unique to each user? I.e. this should just default to 1 for every attempt because it's the first attempt for a user on a given exam?

alangsto: Also should we be setting this value on attempt creation? The first example is giving me pause.

MichaelRoytman: That's what it sounds like to me. It's the combination of the `resource_link_id` (some unique identifier for the exam - presumably the `resource_id` model field on the exam), `sub`, and `attempt_number` claims that uniquely identify an attempt in the LTI launch.

We could set up a composite primary key on (user, exam, attempt_number).

edx_exams/apps/core/api.py
```diff
@@ -142,3 +143,58 @@ def get_exam_attempt_time_remaining(exam_attempt, now=None):
...
+
+    exam_obj = Exam.get_exam_by_id(exam_id)
+
+    if exam_obj is None:
+        err_msg = (
+            f'Exam with exam_id={exam_id} does not exist.'
+        )
+        raise ExamDoesNotExist(err_msg)
+
+    log.info(
```
varshamenon4, line 162: Nit: Was there a particular reason you'd wanted to log before and after creating the exam attempt? The reason I ask is because I wonder if it is confusing for logging if there is a log message but then the attempt is not actually created. But then again, I can see that logging could be useful for debugging purposes and I don't think there's any cost to logging so would defer to you here! :)

edx_exams/apps/core/tests/test_api.py
```diff
@@ -273,3 +278,94 @@ def test_with_no_attempt(self):
...
+            time_limit_mins=30,
+            due_date=timezone.now() - timedelta(minutes=60)
+        )
+
+        with self.assertRaises(ExamAttemptOnPastDueExam) as exc:
+            create_exam_attempt(passed_due_exam.id, self.user.id)
+
+        self.assertIn('trying to create exam attempt for past due non-practice exam', str(exc.exception))
+
+    @ddt.data(
```
varshamenon4, line 329: Nice!

edx_exams/apps/api/v1/views.py
```diff
@@ -453,3 +469,19 @@ def put(self, request, attempt_id):
             status=status.HTTP_400_BAD_REQUEST,
             data={'detail': f'Unrecognized action "{action}"'}
         )
+
+    def post(self, request):
+        """
+        HTTP POST handler to create exam attempt
+        """
+        should_start_immediately = request.data.get('start_clock', 'false').lower() == 'true'
```
varshamenon4, line 477: Nice, I like how you handled the case where you want to make an exam attempt started.

varshamenon4 (commented):
Looks awesome! Just a few questions and nits. :)

edx_exams/apps/core/api.py
```diff
@@ -142,3 +143,58 @@ def get_exam_attempt_time_remaining(exam_attempt, now=None):
         secs_to_end = (end_time-now).total_seconds()
 
     return secs_to_end
+
+
+def create_exam_attempt(exam_id, user_id):
+    """
+    Creates an exam attempt for user_id against exam_id. There should only
+    be one exam_attempt per user per exam.
```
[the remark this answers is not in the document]
alangsto: Good catch - we're not currently supporting multiple attempts for the same exam. However, a user can have multiple attempts if each attempt is associated with a different exam. I need to add logic to make sure that there is not already an existing attempt, and raise an exception if there is. 

For historical context, edx-proctoring does allow multiple attempts per user per exam, which is why you may have seen that logic/references to that behavior in edx-proctoring.

edx_exams/apps/core/api.py
alangsto: I think there's a tendency to log as much info as is useful for proctoring, because then there is a wealth of information available in splunk when trying to debug something. 

It's useful to log before and after so that it's clear if we tried to create an attempt, and if we were actually successful in creating that attempt.

edx_exams/apps/core/api.py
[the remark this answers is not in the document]
alangsto: Just updated so that an error is thrown if an attempt already exists!

edx_exams/apps/core/api.py
alangsto: Because we don't allow multiple attempts for the same exam right now, wouldn't that mean the attempt_code on creation should always be 1?

I know multiple attempts is out of scope right now, but just trying to think about how to add the right functionality/make it easy to add multiple attempts in the future

edx_exams/apps/core/api.py
varshamenon4: Ah that makes total sense and is a really good point.

edx_exams/apps/api/v1/views.py
```diff
@@ -400,13 +405,24 @@ class ExamAttemptView(ExamsAPIView):
...
+
+    HTTP POST
+    Creates a new attempt based on a provided exam_id
+
+    POST data Parameters
+        'exam_id': The unique identifier for the exam
+        'start_clock': Boolean value representing whether or not the attempt should immediately transition to 'started'
+
+    POST Response Values
+        {'exam_attempt_id': <attempt_id>}: The attempt id of the attempt that was created
```
varshamenon4, line 417: Nice docs!

edx_exams/apps/core/api.py
```diff
@@ -142,3 +148,69 @@ def get_exam_attempt_time_remaining(exam_attempt, now=None):
...
+
+    exam_obj = Exam.get_exam_by_id(exam_id)
+
+    if exam_obj is None:
+        err_msg = (
+            f'Exam with exam_id={exam_id} does not exist.'
+        )
+        raise ExamDoesNotExist(err_msg)
+
+    if ExamAttempt.get_current_exam_attempt(user_id, exam_id) is not None:
```
varshamenon4, line 167: Nice check here!

varshamenon4 (approved):
Awesome!

PR #78, opened 2023-02-23
feat: add endpoint to retrieve exam

alangsto:
**JIRA:** [MST-1653](https://2u-internal.atlassian.net/browse/MST-1653)

**Description:** This PR adds an endpoint to return a serialized exam object. If a user has an attempt for that exam, the attempt will also be returned as part of the exam object.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**


edx_exams/apps/api/v1/urls.py
```diff
@@ -32,4 +33,7 @@
     path('exams/attempt',
          ExamAttemptView.as_view(),
          name='exams-attempt',),
+    re_path(fr'exams/attempt/course_id/{COURSE_ID_PATTERN}/content_id/{CONTENT_ID_PATTERN}',
```
alangsto, line 36: In the edx-proctoring url path, the content_id was a query parameter. The MFE always appends the parameter so I decided to make it a path parameter for edx-exams

edx_exams/apps/api/v1/views.py
```diff
@@ -485,3 +488,49 @@ def post(self, request):
...
+        exam = get_exam_by_content_id(course_id, content_id)
+
+        if exam is None:
+            data = {'exam': {}}
+            return Response(data)
+
+        exam_type_class = get_exam_type(exam['exam_type'])
+
+        # the following are additional fields that the frontend expects
+        exam['type'] = exam['exam_type']
```
alangsto, line 523: I'm thinking it might be better to create a custom exam serializer as well, instead of adding these fields here. Any thoughts?

edx_exams/apps/api/serializers.py
```diff
@@ -94,3 +95,56 @@ class Meta:
...
+        return obj.exam.exam_type
+
+    def get_exam_display_name(self, obj):
+        return obj.exam.exam_name
+
+    def get_exam_url_path(self, obj):
+        """
+        Getter for custom exam_url_path field
+        """
+        from edx_exams.apps.core.api import get_exam_url_path  # pylint: disable=import-outside-toplevel
```
alangsto, line 129: Putting the import here to avoid a cyclical import error (which is still happening). Should these be moved to a utils file instead?

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -980,3 +982,96 @@ def test_post_create_attempt(self, start_immediately, mock_update_attempt, mock_
...
+        serialized_exam['type'] = serialized_exam['exam_type']
+        serialized_exam['is_proctored'] = exam_type_class.is_proctored
+        serialized_exam['is_practice_exam'] = exam_type_class.is_practice
+        serialized_exam['backend'] = serialized_exam['provider_name']
+        serialized_exam['attempt'] = {}
+
+        response = self.get_api(self.user, self.course_id, self.content_id)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIsNotNone(response.data['exam'])
```
alangsto, line 1054: need to do more assertions here on the exam object depending on whether or not the exam gets a custom serializer

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -980,3 +982,96 @@ def test_post_create_attempt(self, start_immediately, mock_update_attempt, mock_
...
+            exam=self.exam,
+            attempt_number=1,
+            status=ExamAttemptStatus.created,
+        )
+
+        response = self.get_api(self.user, self.course_id, self.content_id)
+
+        self.assertEqual(response.status_code, 200)
+
+        self.assertIsNotNone(response.data['exam']['attempt'])
```
alangsto, line 1077: I think this assertion is enough? Mostly just want to check that if an attempt exists, we hit `get_student_exam_attempt` and the result of that function is included within the `exam` response object

edx_exams/apps/core/api.py
```diff
@@ -214,3 +216,46 @@ def create_exam_attempt(exam_id, user_id):
...
     return attempt.id
+
+
+def get_exam_by_content_id(course_id, content_id):
+    """
+    Retrieve an exam filtered by a course_id and content_id
+    """
+    try:
+        exam = Exam.objects.get(course_id=course_id, content_id=content_id)
+        return ExamSerializer(exam).data
```
Zacharis278, line 227: Navigating circular dependencies is a sure sign something is off. I think the root of that is that the core app should not be importing from the API. Personally, I don't think serializers belong in this file or the core app at all and we should be doing that in the view.

edx_exams/apps/api/v1/views.py
Zacharis278: eh could go either way on this. Since we're building up a specific response that isn't based on just one model and isn't going to be reused it might be easier to read if the full response is just constructed here

edx_exams/apps/core/api.py
```diff
@@ -214,3 +216,46 @@ def create_exam_attempt(exam_id, user_id):
...
+    Retrieve an attempt with additional data used by the frontend learner view
+    """
+    try:
+        attempt_obj = ExamAttempt.objects.get(id=attempt_id)
+        return StudentAttemptSerializer(attempt_obj).data
+    except ExamAttempt.DoesNotExist:
+        return None
+
+
+def get_exam_url_path(course_id, content_id):
```
Zacharis278, line 254: see earlier comment, this seems like it belongs under the api app.

Zacharis278 (commented):
some initial comments, haven't looked at the tests at all yet

edx_exams/apps/api/serializers.py
```diff
@@ -41,6 +51,7 @@ class ExamSerializer(serializers.ModelSerializer):
     exam_type = serializers.CharField(required=True)
     hide_after_due = serializers.BooleanField(required=True)
     is_active = serializers.BooleanField(required=True)
+    provider_name = serializers.CharField(required=False, source='provider.name')
```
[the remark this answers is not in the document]
alangsto: Yeah, for user facing text. I'll update it!

edx_exams/apps/core/api.py
alangsto: The reason I was adding serializers to the api.py file was to avoid having the view handle any models directly.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -980,3 +982,91 @@ def test_post_create_attempt(self, start_immediately, mock_update_attempt, mock_
...
+
+        headers = self.build_jwt_headers(user)
+        url = reverse(
+            'api:v1:student-course_exam_attempt',
+            kwargs={'course_id': course_id, 'content_id': content_id}
+        )
+
+        return self.client.get(url, **headers)
+
+    def test_no_exam(self):
```
varshamenon4, line 1030: Nice test cases!

edx_exams/apps/api/v1/views.py
```diff
@@ -485,3 +487,51 @@ def post(self, request):
...
+            if any attempt for that exam exists.
+            {
+                'exam': {
+                    'attempt': {...}
+                    ...
+                },
+            }
+    """
+
+    def get(self, request, course_id, content_id):
```
varshamenon4, line 509: The logic of this is really nice and easy to understand!

varshamenon4 (approved):
Awesome!

PR #83, opened 2023-03-06
feat: move exam access token to response body

varshamenon4:
**JIRA:** [MST-1781](https://2u-internal.atlassian.net/browse/MST-1781)

**Description:** This PR moves the exam access token from a cookie to the response body. 
This PR addresses only the first criteria: 
Have the exam service return the token in the request body instead of a cookie.
- Include the expiration as an additional field. This can be used for UI performance improvements if needed later down the line.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -548,14 +548,15 @@ class ExamAccessTokensViewsTests(ExamsAPITestCase):
             is_active=True
         )
         self.exam_id = self.exam.id
-        self.url = reverse('api:v1:exam-access-tokens', kwargs={'exam_id': self.exam_id})
+        self.url = reverse('api:v1:exam-access-tokens',
+                           kwargs={'exam_id': self.exam_id})
 
         self.past_due_date = timezone.now() - timedelta(minutes=5)
         self.past_due_exam = Exam.objects.create(
             resource_id=str(uuid.uuid4()),
             course_id=self.course_id,
             provider=self.test_provider,
-            content_id='abcd1234',
+            content_id='past1234',
             exam_name='test_exam',
             exam_type='proctored',
             time_limit_mins=30,
@@ -575,7 +576,7 @@ class ExamAccessTokensViewsTests(ExamsAPITestCase):
         return self.client.get(url, **headers)
 
     def assert_valid_exam_access_token(self, response, user, exam):
-        token = response.cookies["exam_access_token"].value
+        token = response.data.get("exam_access_token")
         self.assertEqual(unpack_token_for(token, user.lms_user_id).get('course_id'), exam.course_id)
         self.assertEqual(unpack_token_for(token, user.lms_user_id).get('content_id'), exam.content_id)
 
@@ -591,7 +592,7 @@ class ExamAccessTokensViewsTests(ExamsAPITestCase):
         """
         Verify the endpoint returns 404 if exam is not found
         """
-        url = reverse('api:v1:exam-access-tokens', kwargs={'exam_id': 674})
+        url = reverse('api:v1:exam-access-tokens', kwargs={'exam_id': 86})
 
         headers = self.build_jwt_headers(self.user)
         response = self.client.get(url, **headers)
@@ -668,7 +669,7 @@ class ExamAccessTokensViewsTests(ExamsAPITestCase):
             resource_id=str(uuid.uuid4()),
             course_id=self.course_id,
             provider=self.test_provider,
-            content_id='abcd1234',
+            content_id='noduedate234',
             exam_name='test_exam',
             exam_type='proctored',
             time_limit_mins=30,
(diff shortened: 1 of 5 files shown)


edx_exams/apps/api/v1/views.py
```diff
@@ -358,28 +358,25 @@ def get_response(cls, exam, user):
...
+    def get(self, request, course_id, content_id):
         """
         Get exam access token as JWT added as cookie to response.
 
         Exam access token corresponds to given exam and user in request.
         """
-        try:
-            exam = Exam.objects.get(id=exam_id)
-        except ObjectDoesNotExist:
+        exam = get_exam_by_content_id(course_id, content_id)
```
varshamenon4, line 377: [Question] I noticed while creating unit tests that get_exam_by_content_id can return multiple exams if the content_ids are duplicates. Is this being checked/prevented somewhere else? In either case, would it be better to have a check in this method to make sure that there aren't duplicates? And to handle the error if so?

edx_exams/apps/api/v1/views.py
```diff
@@ -358,28 +358,25 @@ def get_response(cls, exam, user):
...
-                            data=data)
-
         if grant_access:
             log.info("Creating exam access token")
             access_token = sign_token_for(user.lms_user_id, expiration_window, claims)
-            response.set_cookie(
-                settings.ACCESS_TOKEN_COOKIE_NAME, access_token,
-                domain=settings.ACCESS_TOKEN_COOKIE_DOMAIN
-            )
+            data = {ACCESS_TOKEN_COOKIE_NAME: access_token, ACCESS_TOKEN_COOKIE_NAME + "_expiration": expiration_window}
```
varshamenon4, line 364: Added both access token and expiration window to the body of the response. Let me know if there is a better way to do this or if you think the expiration field should be renamed.

edx_exams/apps/api/v1/views.py
```diff
@@ -358,28 +358,25 @@ def get_response(cls, exam, user):
         elif exam.due_date is not None and timezone.now() >= exam.due_date:
             data, grant_access, response_status = {}, True, status.HTTP_200_OK
```
varshamenon4, line 359: todo: I realized it's probably best to remove all the setting of data to {} here and for any cases where access is granted. I can change to not set here and only set below if access is granted.

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -787,6 +794,38 @@ def test_access_granted_past_due_exam_no_attempt(self):
         self.assertEqual(200, response.status_code)
         self.assert_valid_exam_access_token(response, self.user, self.past_due_exam)
 
+    @ddt.data(
+        (timedelta(minutes=20), timedelta(minutes=0), 200, 60),  # exam attempt time remaining more than default exp
+        # todo: calculate and change this to be testing less than default exp
```
varshamenon4, line 799: Todo: I need to calculate and reset the values here to test that the expiration window is correct when the expiration window is set to be different than the default (when the time remaining is less than the default exp window). That being said, let me know if this test approach makes sense! Thanks.

varshamenon4 (commented):
Added some specific questions and todos for myself. Thanks for the feedback!

Zacharis278 (commented):
few comments, still need to look at the test files

edx_exams/apps/api/v1/views.py
```diff
@@ -358,28 +358,25 @@ def get_response(cls, exam, user):
...
         Get exam access token as JWT added as cookie to response.
 
         Exam access token corresponds to given exam and user in request.
         """
-        try:
-            exam = Exam.objects.get(id=exam_id)
-        except ObjectDoesNotExist:
+        exam = get_exam_by_content_id(course_id, content_id)
+
+        if exam is None:
```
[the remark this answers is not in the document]
varshamenon4: I was thinking of including this as a check before making the fetch call in frontend-app-learning, but this does make me wonder if this is actually possible to check at the unit level. So let me look into that.

Let me know if you have any thoughts or sense of this off the top of your head. Thanks!

edx_exams/apps/api/v1/views.py
varshamenon4: Ah okay. And to clarify, is it possible to have duplicates by content_id? Or is this check not needed?

Zacharis278: it shouldn't be possible to create duplicates so maybe we want a `unique_together` constraint on (course_id, content_id, is_active).

varshamenon4: That sounds good!

edx_exams/apps/api/v1/views.py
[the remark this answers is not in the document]
varshamenon4: Will handle this in the frontend.

edx_exams/apps/api/v1/views.py
varshamenon4: Since we are keeping the old endpoint now, I will create a separate bug ticket to fix this bug

Zacharis278 (approved):
lgtm 👍

PR #88, opened 2023-03-10
[DRAFT] Middleware to handle brokering UI requests to edx-proctoring

Zacharis278:
**JIRA:** [MST-1654](https://2u-internal.atlassian.net/browse/MST-1654

**Description:** 
There are 4 endpoints the UI calls that require some logic to determine if they should look at edx-exams or edx-proctoring. This PR will refer to them by the UI function names here https://github.com/openedx/frontend-lib-special-exams/blob/main/src/data/api.js. The original goal was to have ALL requests go through edx-exams and simply re-route based on the proctoring configuration for the course. I discovered an issue with this approach. While requests for a specific exam or attempt don't always have the course id in context it's pretty trivial to find based on attempt -> exam -> course. _However_, those exam/attempt ids are not guaranteed to be unique between edx-proctoring and edx-exams. Theres a chance that an attempt with an id of 123 exists in both services for two different courses. If that happens the intent of the request becomes ambiguous.

The proposed solution here is to have edx-exams route the initial request for exam state that is loaded in the UI. Followup requests for a specific exam however will just use a field returned by the initial fetch to know what service to hit. So the UI is doing some switching logic but it's still ultimately driven by a decision made on the backend.

`fetchExamAttemptData` This is the first request made by the UI to initialize state, the UI will always hit edx-exams for this. The request includes the course id and will be proxied by the middleware to edx-proctoring if necessary depending on the provider configuration for that course. Responses from edx-proctoring will include a field indicating the legacy service should be used for future calls to create/update attempts for this exam. (this requires a 1-line change to edx-proctoring to add)

`createExamAttempt` The UI will use the field from `fetchExamAttemptData` to determine what service to make this request against. No middleware routing for this endpoint.

`updateAttemptStatus` The UI will use the field from `fetchExamAttemptData` to determine what service to make this request against. No middleware routing for this endpoint.

`pollExamAttempt` (latest/active attempt) still working this out solution TDB

**Author concerns:** minor changes to edx-proctoring are needed to support this


edx_exams/apps/router/interop.py
```diff
@@ -23,10 +29,58 @@ def register_exams(course_id, exam_list):
...
+        )
+
+    return response_data, response.status_code
+
+def get_student_exam_attempt_data(course_id, content_id, user_id):
+    """
+    Get student exam attempt data from the legacy proctoring service
+    """
+    content_id_url_safe = quote_plus(content_id)    # because this goes in the query string
+    path = LMS_PROCTORED_EXAM_ATTEMPT_DATA_API_TPL.format(course_id, content_id_url_safe, user_id)
```
Zacharis278, line 47: The addition of the user_id here requires a minor enhancement to edx-proctoring. Right now this endpoint is only able to look at request.user to determine what user's exam data to get. Since we're tunneling this through edx-exams the user becomes the exams service worker and not the learner.

edx-proctoring needs to support a user_id param that can be used by staff or system users to view another user's exam state.

edx_exams/apps/router/interop.py
```diff
@@ -23,10 +29,58 @@ def register_exams(course_id, exam_list):
         response = make_request('PATCH', url, client, json=exam_list)
     except HTTPError as e:
         response = e.response
-    return response
 
+    response_data = _get_json_data(response)
```
Zacharis278, line 33: the LMS doesn't always return JSON when things go wrong this makes a mess of things on our end. Wrap all these responses so we aren't logging or returning junk html.

edx_exams/apps/router/interop.py
```diff
@@ -1,15 +1,24 @@
...
 from django.conf import settings
 from requests.exceptions import HTTPError
+from rest_framework import status
+from simplejson import JSONDecodeError
 
 from edx_exams.apps.core.rest_utils import get_client, make_request
 
 LMS_PROCTORING_PLUGIN_BASE_PATH = 'api/edx_proctoring/v1/'
 LMS_REGISTER_PROCTORED_EXAMS_API_TPL = 'proctored_exam/exam_registration/course_id/{}'
+LMS_PROCTORED_EXAM_ACTIVE_ATTEMPT_API_TPL = 'proctored_exam/active_attempt?user_id={}'
```
Zacharis278, line 17: this endpoint to just return the 'current' attempt does not exist but is just a few lines of copy/paste from the existing `proctored_exam/attempt/course_id/{}` to set it up. I'll have a PR for that to follow.

edx_exams/apps/router/interop.py
alangsto: Nice, really like this enhancement

edx_exams/apps/router/views.py
```diff
@@ -40,14 +39,31 @@ def patch(self, request, course_id):
...
+            safe=False,
+        )
+
+
+class CourseExamAttemptLegacyView(APIView):
+    """
+    View to handle attempts for exams managed by edx-proctoring.
+    """
+    authentication_classes = (JwtAuthentication,)
+    permission_classes = (StaffUserOrReadOnlyPermissions,)
```
Zacharis278, line 56: note to self, this can just use IsAuthenticated since there is only a GET

edx_exams/apps/router/interop.py
```diff
@@ -23,10 +32,67 @@ def register_exams(course_id, exam_list):
...
+    content_id_url_safe = quote_plus(content_id)    # because this goes in the query string
+    path = LMS_PROCTORED_EXAM_ATTEMPT_DATA_API_TPL.format(course_id, content_id_url_safe, lms_user_id)
+    url = _proctoring_api_url(path)
+    client = get_client(settings.LMS_ROOT_URL)
+    try:
+        response = make_request('GET', url, client)
+    except HTTPError as e:
+        response = e.response
+
+    response_data = _get_json_data(response)
```
alangsto, line 59: I don't know if this is logging overkill, but it might be helpful to log that exams is getting the attempt data from the legacy service (just for debugging purposes in the future).

I'm not sure what edx-proctoring logs though

alangsto (approved):
Left one question about logging, but looks good 👍

edx_exams/apps/router/interop.py
Zacharis278: oh there's a log for this up in the middleware that says the request will be forwarded.

alangsto: whoops sorry for missing that. sounds good 👍

PR #94, opened 2023-03-23
feat: simplify exam publish

Zacharis278:
**JIRA:** [MST-1581](https://2u-internal.atlassian.net/browse/MST-1581)

**Description:** Simplifies the exam publish behavior. The original intent was to allow course teams to change exam types from proctored>timed>proctored and end up with the same proctored exam instance throughout. We found this added some unforeseen complexities and we choose not to support it. (see discovery in linked ticket)

Instead, when an exam type changes we create an entirely new version of that exam.


alangsto (approved):
nice 👍

PR #99, opened 2023-04-06
feat: update exam attempt status when starting LTI proctoring launch

MichaelRoytman:
**JIRA:** [MST-1842](https://2u-internal.atlassian.net/browse/MST-1842)

**Description:** 

This commits adds behavior to update a learner's exam attempt status to the "download_software_clicked" status when the learner starts the LTI proctoring launch via the start_proctoring view.

**Author concerns:** 

* The `start_proctoring` view only handles `GET` requests. This doesn't feel as appropriate given that it changes the `ExamAttempt` state - it's a side effecting GET. However, the URL doesn't really uniquely define a resource (i.e. the `ExamAttempt`). A `PUT` would be more appropriate, but it would make it more difficult to manually test without wiring up the learning MFE to `edx-exams` because you can no longer just put the URL in the browser. What do you think? I can log a ticket to update the request handling once we have a better way to test the entire LTI flow via the learning MFE.

**Dependencies:** None.

**Installation instructions:** Please follow the instructions in [Local Development and LTI Configuration](https://2u-internal.atlassian.net/wiki/spaces/PT/pages/256737327/Local+Development+and+LTI+Configuration) to set up your environment.

**Testing instructions:**

1. Create an `ExamAttempt` with the `created` status.
2. Visit `http://localhost:18740/lti/start_proctoring/<attempt_id>`.
3. Verify that the `ExamAttempt` has transitioned to the `download_software_clicked` state and that you see the corresponding logs in the console.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**


edx_exams/apps/lti/views.py
```diff
@@ -69,8 +91,8 @@ def end_assessment(request, attempt_id):
     LTI End Assessment
     """
     # TODO: Here we'd do all the end of assessment things.
-    exam_attempt = ExamAttempt.objects.get(pk=attempt_id)
-    exam = exam_attempt.exam
+    attempt = ExamAttempt.objects.get(pk=attempt_id)
+    exam = attempt.exam
```
MichaelRoytman, line 95: I changed the variable names to be more consistent with the rest of the application, so I'm going to ignore the Codecov warning for now.

edx_exams/apps/lti/views.py
```diff
@@ -13,31 +13,53 @@
...
 
+from edx_exams.apps.core.api import get_attempt_by_id, update_attempt_status
 from edx_exams.apps.core.models import ExamAttempt
+from edx_exams.apps.core.statuses import ExamAttemptStatus
 from edx_exams.apps.lti.utils import get_lti_root
 
 EDX_OAUTH_BACKEND = 'auth_backends.backends.EdXOAuth2'
 
 
+@api_view(['GET'])
```
MichaelRoytman, line 29: The start_proctoring view only handles GET requests. This doesn't feel as appropriate given that it changes the ExamAttempt state - it's a side effecting GET. However, the URL doesn't really uniquely define a resource (i.e. the ExamAttempt). A PUT would be more appropriate, but it would make it more difficult to manually test without wiring up the learning MFE to edx-exams because you can no longer just put the URL in the browser. What do you think? I can log a ticket to update the request handling once we have a better way to test the entire LTI flow via the learning MFE.

Zacharis278: I think this has to be a GET because it isn't really an 'API request' and instead a new webpage the browser links to. Not sure how to make the side effect here more elegant 🤔

Zacharis278: maybe we make sure it is idempotent-ish so in case the learner hits the same launch URL twice it doesn't error, it simply always results in download_software_clicked? (if that is a valid transition)

edx_exams/apps/lti/views.py
Zacharis278: can you just pragma: no-cover this whole function?

edx_exams/apps/lti/views.py
```diff
@@ -13,31 +13,53 @@
...
 @permission_classes((IsAuthenticated,))
 def start_proctoring(request, attempt_id):
     """
     LTI Start Proctoring
     """
-    # TODO: Here we'd do all the start of proctoring things.
+    # Verify that the attempt associated with attempt_id exists.
+    attempt = get_attempt_by_id(attempt_id)
+    if not attempt:
+        return Response(
```
Zacharis278, line 40: something to follow up with as a 'nice to have' ticket but these are going to end up rendered in the user's browser. We might want to return a nicer HTML fragment or something.

edx_exams/apps/lti/views.py
```diff
@@ -13,31 +13,53 @@
...
+    user = request.user
+    if attempt.user.id != user.id:
+        error_msg = (
+            f"user_id={attempt.user.id} attempted to update attempt_id={attempt.id} in "
+            f"course_id={attempt.exam.course_id} when starting an LTI launch but does not have access to it."
+        )
+        error = {'detail': error_msg}
+        return Response(status=status.HTTP_403_FORBIDDEN, data=error)
+
+    update_attempt_status(attempt_id, ExamAttemptStatus.download_software_clicked)
```
Zacharis278, line 55: I think we should catch and handle any illegal status transitions here

edx_exams/apps/lti/views.py
MichaelRoytman: Your second thought about the auto-submitting form is separate from the GET v. POST topic, right? I'm not seeing the connection. What do you mean by "more correct"? The specification does say a redirect is acceptable, so I'm trying to understand the benefit/difference.

MichaelRoytman: When I've done multiple launch in succession, I haven't had any errors in the state transitions, so it appears idempotent.

edx_exams/apps/lti/views.py
MichaelRoytman: That's a good point. I'll log a ticket.

edx_exams/apps/lti/views.py
MichaelRoytman: I could, but I was a little concerned we'd forget to remove it. I'll add a comment to https://2u-internal.atlassian.net/browse/MST-1845 to remove it once we add tests for this view.

edx_exams/apps/lti/views.py
Zacharis278: I was thinking 'more correct' in terms of HTTP specification so that our launch url becomes a POST. But, for now at least I see no reason that extra complexity is worth it.

edx_exams/apps/lti/views.py
MichaelRoytman: I logged [MST-1865](https://2u-internal.atlassian.net/browse/MST-1865).

edx_exams/apps/lti/views.py
MichaelRoytman: Ah, I see. I thought you meant replace the LTI redirect here with an auto-posting form, which is what we do in the LMS, actually. That's where my confusion was.

Zacharis278 (approved):
lgtm

PR #107, opened 2023-04-11
feat: update exam attempt status after LTI_1P3_PROCTORING_ASSESSMENT_…

MichaelRoytman:
**JIRA:** [MST-1843](https://2u-internal.atlassian.net/browse/MST-1843)

**Description:** This commits adds behavior to update a learner's exam attempt status to the `ready_to_start` status when the `LTI_1P3_PROCTORING_ASSESSMENT_STARTED` signal is received.

**Author concerns:** None.
**Dependencies:** None.

**Installation instructions:** Please follow the instructions in [Local Development and LTI Configuration](https://2u-internal.atlassian.net/wiki/spaces/PT/pages/256737327/Local+Development+and+LTI+Configuration) to set up your environment.

**Testing instructions:**

**Testing instructions:**

1. Create an `ExamAttempt` with the `created` status.
2. Open the Django shell from the command line. `python manage.py shell`
3. Import and run the signal.
```
from django.contrib.auth import get_user_model
from lti_consumer.signals.signals import LTI_1P3_ASSESSMENT_STARTED
from edx_exams.apps.core.models import Exam, ExamAttempt

user_model = get_user_model()
user = user_model.objects.get(username='edx')

exam = Exam.objects.get(id=<id>)
attempt_number = ExamAttempt.objects.get(id=<id>

LTI_1P3_PROCTORING_ASSESSMENT_STARTED.send(sender=None, attempt_number=attempt.attempt_number, resource_id=exam.resource_id, user_id=user.id)
```
4. Verify that the `ExamAttempt` has transitioned to the `ready_to_start` state and that you see the corresponding logs in the console.
5. Create additional `ExamAttempts` to verify error cases by re-sending the signal with the appropriate arguments.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/lti/signals/handlers.py
+++ b/edx_exams/apps/lti/signals/handlers.py
@@ -1,8 +1,16 @@
 """
 edX Exams Signal Handlers
 """
+import logging
+
 from django.dispatch import receiver
 from lti_consumer.signals.signals import LTI_1P3_PROCTORING_ASSESSMENT_STARTED
+
+from edx_exams.apps.core.api import get_attempt_for_user_with_attempt_number_and_resource_id, update_attempt_status
+from edx_exams.apps.core.exceptions import ExamIllegalStatusTransition
+from edx_exams.apps.core.statuses import ExamAttemptStatus
+
+log = logging.getLogger(__name__)
 
 
 @receiver(LTI_1P3_PROCTORING_ASSESSMENT_STARTED)
@@ -10,4 +18,36 @@ def assessment_started(sender, **kwargs):  # pylint: disable=unused-argument
     """
     Signal handler for the lti_consumer LTI_1P3_PROCTORING_ASSESSMENT_STARTED signal.
     """
-    print(f"LTI_1P3_PROCTORING_ASSESSMENT_STARTED signal received with kwargs: {kwargs}")
+    user_id = kwargs.get('user_id')
+    attempt_number = kwargs.get('attempt_number')
+    resource_id = kwargs.get('resource_id')
+
+    if not user_id or not attempt_number or not resource_id:
+        log.info(
+            'expected valid values for LTI_1P3_PROCTORING_ASSESSMENT_STARTED signal kwargs but received '
+            f'attempt_number={attempt_number}, user_id={user_id}, and resource_id={resource_id}.'
+        )
+        return
+
+    attempt = get_attempt_for_user_with_attempt_number_and_resource_id(
+        user_id,
+        attempt_number,
+        resource_id,
+    )
+
+    if not attempt:
+        log.info(
+            f'attempt_number={attempt_number} for user_id={user_id} in exam with resource_id={resource_id} is '
+            'not associated with a single attempt.'
+        )
+        return
+
+    try:
+        update_attempt_status(attempt.id, ExamAttemptStatus.ready_to_start)
+    except ExamIllegalStatusTransition:
+        log.info(
+            f'user_id={attempt.user.id} attempted to update attempt_id={attempt.id} in '
+            f'course_id={attempt.exam.course_id} from status={attempt.status} to '
+            f'status={ExamAttemptStatus.ready_to_start}, but this status '
+            f'transition is illegal.'
+        )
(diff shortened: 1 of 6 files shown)


edx_exams/apps/lti/signals/tests/test_handlers.py
```diff
@@ -0,0 +1,142 @@
...
+        self.attempt = ExamAttempt.objects.create(
+            user=self.user,
+            exam=self.exam,
+            attempt_number=1111111,
+            status=ExamAttemptStatus.created,
+            start_time=None,
+            allowed_time_limit_mins=None,
+        )
+
+    def test_assessment_started_updated_attempt(self):
```
alangsto, line 60: Is there a reason that `update_attempt_status` isn't being mocked here? That function already has tests, so instead of testing the whole code path here you could check that the mock has been called

edx_exams/apps/lti/signals/tests/test_handlers.py
```diff
@@ -0,0 +1,142 @@
...
+        assessment_started(None, **kwargs)
+
+        self.attempt.refresh_from_db()
+
+        # Both attempt's attempt status should remain unchanged because the kwargs refer to more than one attempt.
+        assert not mock_update_attempt_status.called
+        self.assertEqual(self.attempt.status, ExamAttemptStatus.created)
+        self.assertEqual(other_attempt.status, ExamAttemptStatus.created)
+
+    def test_assessment_started_illegal_transition(self):
```
alangsto, line 123: Similar question here, just wondering why `update_attempt_status` isn't mocked with a side effect

alangsto (approved):
Left two questions about mocking for tests, but otherwise looks good!

edx_exams/apps/lti/signals/handlers.py
```diff
@@ -1,13 +1,46 @@
...
+log = logging.getLogger(__name__)
+
 
 @receiver(LTI_1P3_PROCTORING_ASSESSMENT_STARTED)
 def assessment_started(sender, **kwargs):  # pylint: disable=unused-argument
     """
     Signal handler for the lti_consumer LTI_1P3_PROCTORING_ASSESSMENT_STARTED signal.
     """
-    print(f"LTI_1P3_PROCTORING_ASSESSMENT_STARTED signal received with kwargs: {kwargs}")
+    user_id = kwargs.get('user_id')
```
Zacharis278, line 21: since the thing that sends this is in another codebase I'm wondering if we should more explicitly check these args are correctly provided? I think the log under `if not attempt:` technically covers this but could leave us open to a future error if we're expecting these not to be null in an underlying function of `get_attempt_for_user_with_attempt_number_and_resource_id`

MichaelRoytman: Yeah, I was thinking about this and went back and forth a bit. I figured we own the LTI library so it would be pretty safe, but it doesn't hurt to be extra careful and not assume anything about the signal, so I'll check them here.

edx_exams/apps/lti/signals/tests/test_handlers.py
MichaelRoytman: I don't typically mock unless it's absolutely necessary. I had to mock in the some of these other tests because I'm testing that nothing happened, and that's hard to do without asserting on _something_, so I chose to ensure that `update_attempt_status` wasn't called. I'm open to alternatives to avoid mocking in these tests!

Otherwise, mocking out that function feels like testing an implementation detail. I'd like to test the result of calling the handler, not necessarily how the handler achieves that result. In my opinion, the test shouldn't care how the status was updated - just that it was updated somehow. If I mock out `update_attempt_status`, then these tests are reliant on the assumption that `update_attempt_status` is called in the function, and they will break anytime we change the way we update attempt statuses or change the function name, making them brittle. If we don't mock out the function, the tests will continue to function regardless.

What do you think?

alangsto: That makes sense!

edx_exams/apps/lti/signals/handlers.py
MichaelRoytman: @Zacharis278 I added some checking below. What do you think? Let me know if you'd rather I explicitly check for `None`.

Zacharis278: looks good!

Zacharis278 (approved):

PR #120, opened 2023-04-24
temp: POC for importing lti auth/perm classes

ilee2u:
Temporary PR not meant to be merged to fulfill [MST-1885](https://2u-internal.atlassian.net/browse/MST-1885)


ilee2u:
Note that I had to install xblock-lti-consumer as an editable, which I believe is why pytest is currently throwing errors.

PR #124, opened 2023-05-03
feat: acs endpoint

ilee2u:
**JIRA:** [Link to JIRA ticket](https://2u-internal.atlassian.net/browse/MST-1847)

**Description:** Adds an ACS endpoint to /lti/views.py

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/core/api.py
+++ b/edx_exams/apps/core/api.py
@@ -48,6 +48,18 @@ def get_latest_attempt_for_user(user_id):
     return latest_attempt
 
 
+def get_user_by_anonymous_id(anonymous_user_id, attempt_number, resource_id):
+    """
+    Get an exam attempt described by anonymous_user_id, resource_id, and the attempt_number.
+    """
+    attempt = ExamAttempt.objects.get(
+        user__anonymous_user_id=anonymous_user_id,
+        attempt_number=attempt_number,
+        exam__resource_id=resource_id
+    )
+    return attempt
+
+
 def get_attempt_for_user_with_attempt_number_and_resource_id(user_id, attempt_number, resource_id):
     """
     Retrieve an attempt in an exam described by resource_id for a user described by user_id with a particular attempt
--- a/edx_exams/apps/core/api.py
+++ b/edx_exams/apps/core/api.py
@@ -48,17 +48,6 @@ def get_latest_attempt_for_user(user_id):
     return latest_attempt
 
 
-def get_attempt_by_attempt_number_and_resource_id(attempt_number, resource_id):
-    """
-    Retrieve an attempt in an exam described by resource_id with a particular attempt
-    number described by attempt_number.
-    """
-    return ExamAttempt.get_attempt_by_attempt_number_and_resource_id(
-        attempt_number,
-        resource_id,
-    )
-
-
 def get_attempt_for_user_with_attempt_number_and_resource_id(user_id, attempt_number, resource_id):
     """
     Retrieve an attempt in an exam described by resource_id for a user described by user_id with a particular attempt
--- a/edx_exams/apps/core/api.py
+++ b/edx_exams/apps/core/api.py
@@ -48,6 +48,17 @@ def get_latest_attempt_for_user(user_id):
     return latest_attempt
 
 
+def get_attempt_by_attempt_number_and_resource_id(attempt_number, resource_id):
+    """
+    Retrieve an attempt in an exam described by resource_id with a particular attempt
+    number described by attempt_number.
+    """
+    return ExamAttempt.get_attempt_by_attempt_number_and_resource_id(
+        attempt_number,
+        resource_id,
+    )
+
+
 def get_attempt_for_user_with_attempt_number_and_resource_id(user_id, attempt_number, resource_id):
     """
     Retrieve an attempt in an exam described by resource_id for a user described by user_id with a particular attempt
--- a/edx_exams/apps/core/api.py
+++ b/edx_exams/apps/core/api.py
@@ -52,11 +52,20 @@ def get_user_by_anonymous_id(anonymous_user_id, attempt_number, resource_id):
     """
     Get an exam attempt described by anonymous_user_id, resource_id, and the attempt_number.
     """
-    attempt = ExamAttempt.objects.get(
-        user__anonymous_user_id=anonymous_user_id,
-        attempt_number=attempt_number,
-        exam__resource_id=resource_id
-    )
+    try:
+        attempt = ExamAttempt.objects.get(
+            user__anonymous_user_id=anonymous_user_id,
+            attempt_number=attempt_number,
+            exam__resource_id=resource_id
+        )
+    except ExamAttempt.DoesNotExist:
+        return None
+    except ExamAttempt.MultipleObjectsReturned:
+        log.warning(
+            f'attempt_number={attempt_number} for anonymous user id={anonymous_user_id} '
+            f'in exam with resource_id={resource_id} is associated with multiple attempts.'
+        )
+        return None
     return attempt
 
 
--- a/edx_exams/apps/core/api.py
+++ b/edx_exams/apps/core/api.py
@@ -48,27 +48,6 @@ def get_latest_attempt_for_user(user_id):
     return latest_attempt
 
 
-def get_user_by_anonymous_id(anonymous_user_id, attempt_number, resource_id):
-    """
-    Get an exam attempt described by anonymous_user_id, resource_id, and the attempt_number.
-    """
-    try:
-        attempt = ExamAttempt.objects.get(
-            user__anonymous_user_id=anonymous_user_id,
-            attempt_number=attempt_number,
-            exam__resource_id=resource_id
-        )
-    except ExamAttempt.DoesNotExist:
-        return None
-    except ExamAttempt.MultipleObjectsReturned:
-        log.warning(
-            f'attempt_number={attempt_number} for anonymous user id={anonymous_user_id} '
-            f'in exam with resource_id={resource_id} is associated with multiple attempts.'
-        )
-        return None
-    return attempt
-
-
 def get_attempt_for_user_with_attempt_number_and_resource_id(user_id, attempt_number, resource_id):
     """
     Retrieve an attempt in an exam described by resource_id for a user described by user_id with a particular attempt
--- a/edx_exams/apps/core/models.py
+++ b/edx_exams/apps/core/models.py
@@ -232,7 +232,6 @@ class ExamAttempt(TimeStampedModel):
         except cls.DoesNotExist:
             return True
 
-
     @classmethod
     def get_attempt_for_user_with_attempt_number_and_resource_id(cls, user_id, attempt_number, resource_id):
         """
@@ -240,7 +239,6 @@ class ExamAttempt(TimeStampedModel):
         attempt number described by attempt_number.
         """
         try:
-            print("User ID is int is:", isinstance(user_id, int))
             # If user_id is an integer, use our internal database ID
             if isinstance(user_id, int):
                 attempt = ExamAttempt.objects.get(
--- a/edx_exams/apps/core/models.py
+++ b/edx_exams/apps/core/models.py
@@ -240,11 +240,21 @@ class ExamAttempt(TimeStampedModel):
         attempt number described by attempt_number.
         """
         try:
-            attempt = ExamAttempt.objects.get(
-                user_id=user_id,
-                attempt_number=attempt_number,
-                exam__resource_id=resource_id
-            )
+            print("User ID is int is:", isinstance(user_id, int))
+            # If user_id is an integer, use our internal database ID
+            if isinstance(user_id, int):
+                attempt = ExamAttempt.objects.get(
+                    user_id=user_id,
+                    attempt_number=attempt_number,
+                    exam__resource_id=resource_id
+                )
+            # If not an integer, use the User object's anonymous_user_id instead
+            else:
+                attempt = ExamAttempt.objects.get(
+                    user__anonymous_user_id=user_id,
+                    attempt_number=attempt_number,
+                    exam__resource_id=resource_id
+                )
             return attempt
         except cls.DoesNotExist:
             return None
--- a/edx_exams/apps/core/models.py
+++ b/edx_exams/apps/core/models.py
@@ -232,26 +232,6 @@ class ExamAttempt(TimeStampedModel):
         except cls.DoesNotExist:
             return True
 
-    @classmethod
-    def get_attempt_by_attempt_number_and_resource_id(cls, attempt_number, resource_id):
-        """
-        Retrieve an attempt in an exam described by resource_id with a particular attempt
-        number described by attempt_number.
-        """
-        try:
-            attempt = ExamAttempt.objects.get(
-                attempt_number=attempt_number,
-                exam__resource_id=resource_id
-            )
-            return attempt
-        except cls.DoesNotExist:
-            return None
-        except cls.MultipleObjectsReturned:
-            log.warning(
-                f'attempt_number={attempt_number} in exam with resource_id={resource_id} is '
-                'associated with multiple attempts.'
-            )
-            return None
 
     @classmethod
     def get_attempt_for_user_with_attempt_number_and_resource_id(cls, user_id, attempt_number, resource_id):
--- a/edx_exams/apps/core/models.py
+++ b/edx_exams/apps/core/models.py
@@ -239,20 +239,11 @@ class ExamAttempt(TimeStampedModel):
         attempt number described by attempt_number.
         """
         try:
-            # If user_id is an integer, use our internal database ID
-            if isinstance(user_id, int):
-                attempt = ExamAttempt.objects.get(
-                    user_id=user_id,
-                    attempt_number=attempt_number,
-                    exam__resource_id=resource_id
-                )
-            # If not an integer, use the User object's anonymous_user_id instead
-            else:
-                attempt = ExamAttempt.objects.get(
-                    user__anonymous_user_id=user_id,
-                    attempt_number=attempt_number,
-                    exam__resource_id=resource_id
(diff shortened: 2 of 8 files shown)


edx_exams/apps/lti/views.py
```diff
@@ -23,8 +28,53 @@
...
+    We can implement proper functionality and tests later once we
+    hear back from our third party proctoring service vendors
+    (i.e. Verificient and Proctortrack).
+    """
+    # TODO:
+    # First, have access_token in lti-consumer add actions to the access token
+    data = json.loads(request.body)
+    action = data['action']
+
+    if action == 'flag':
```
Zacharis278, line 61: I'd recommend doing a bit more than just looking at the action here. We can take the user, resource, and attempt number in this request to look up the attempt object that we would be modifying or flagging. Can we throw some errors if those parameters are incorrect? Maybe make our log message more explicit on exactly what attempt is being flagged?

edx_exams/apps/lti/views.py
```diff
@@ -23,8 +28,53 @@
...
+        log.info('Flagging exam attempt')
+
+    elif action == 'pause':
+        log.info('Pausing exam attempt')
+
+    elif action == 'resume':
+        log.info('Resuming exam attempt')
+
+    elif action == 'terminate':
```
Zacharis278, line 70: since we're telling the tool we only support flag request I don't think we need conditions for all of the other possible actions yet (until we choose to support them)

edx_exams/apps/lti/urls.py
```diff
@@ -8,6 +8,7 @@
 
 app_name = 'lti'
 urlpatterns = [
+    path('acs_endpoint/<int:lti_config_id>', views.acs_endpoint, name='acs_endpoint'),
```
Zacharis278, line 11: NIT: 'endpoint' seems kinda redundant to put in a url. Maybe just acs or assessment_control. I'd also recommend putting the config id first in this url since that id refers to the lti configuration and not the assessment. IMO putting it last implies it's an assessment/attempt id.

so` /lti/<lti_configuration>/acs` or something like that

edx_exams/apps/lti/views.py
```diff
@@ -12,19 +15,96 @@
...
+
+    attempt = get_attempt_by_attempt_number_and_resource_id(attempt_number, resource_id)
+    if attempt.status not in VALID_STATUSES:
+        # TODO: improve this msg with fields and stuff
+        log.info(
+            f'Attempt cannot be flagged for user {user} in',
+            f'with resource id {resource_id} and attempt',
+            f'number {attempt_number}.'
+        )
+        # NOTE: Do we want to create a new exception in exceptions.py just for this case?
```
ilee2u, line 100: Just want to highlight this question about making a new exception for this case.

edx_exams/apps/core/models.py
```diff
@@ -234,6 +234,27 @@ def check_no_other_active_attempts_for_user(cls, user_id, attempt_id):
         except cls.DoesNotExist:
             return True
 
+    @classmethod
+    def get_attempt_by_attempt_number_and_resource_id(cls, attempt_number, resource_id):
```
Zacharis278, line 238: nothing wrong with having this here I just want to point out that we don't need always need this type of logic to be on the model class directly. If it's a pretty specific/narrow use case (filter by resource_id feels pretty narrow) it's totally fine to just put in on api.py. We don't always need the extra layer of abstraction.

My rule of thumb here is to do this logic on the model if I expect it would be used multiple times elsewhere.

ilee2u:
Also, I'm aware that I haven't reached full coverage. I want to make sure my approach for this ticket is okay first before making tests.

edx_exams/apps/core/api.py
```diff
@@ -41,6 +41,17 @@ def get_latest_attempt_for_user(user_id):
     return latest_attempt
 
 
+def get_attempt_by_attempt_number_and_resource_id(attempt_number, resource_id):
```
Zacharis278, line 44: it looks like we have a very similar function already for `get_attempt_for_user_with_attempt_number_and_resource_id` any way we can cut down on duplicate code paths here?

edx_exams/apps/lti/views.py
```diff
@@ -12,19 +15,96 @@
...
+    # But not verified, rejected, or expired (no need to flag in these cases)
+    # Therefore, any attempt with status from 'started' to 'submitted' can be flagged
+    VALID_STATUSES = [
+        ExamAttemptStatus.started,
+        ExamAttemptStatus.ready_to_submit,
+        ExamAttemptStatus.timed_out,
+        ExamAttemptStatus.submitted,
+    ]
+
+    attempt = get_attempt_by_attempt_number_and_resource_id(attempt_number, resource_id)
```
Zacharis278, line 92: this would also need to look at the user id because attempt_number is not unique. It's always 1 for a user's first attempt, 2 for the second, and so on.

Zacharis278: What's our error condition if we can't find an attempt for this user and this returns None?

edx_exams/apps/lti/views.py
```diff
@@ -12,19 +15,96 @@
...
+        log.info(
+            f'Attempt cannot be flagged for user {user} in',
+            f'with resource id {resource_id} and attempt',
+            f'number {attempt_number}.'
+        )
+        # NOTE: Do we want to create a new exception in exceptions.py just for this case?
+        return Response(status=400)
+
+    if action == 'flag':
+        log.info('Flagging exam attempt')
```
Zacharis278, line 104: I think it's worth logging the user_id, exam_id, and attempt_id here

Zacharis278 (commented):
some initial comments

edx_exams/apps/core/models.py
```diff
@@ -241,11 +241,20 @@ def get_attempt_for_user_with_attempt_number_and_resource_id(cls, user_id, attem
         attempt number described by attempt_number.
         """
         try:
-            attempt = ExamAttempt.objects.get(
-                user_id=user_id,
-                attempt_number=attempt_number,
-                exam__resource_id=resource_id
-            )
+            # If user_id is an integer, use our internal database ID
```
ilee2u, line 244: I'm wondering if we should split this logic between two different model functions, with the if/else statement happening in api.py instead.

Zacharis278: tbh I wouldn't hide this behavior all the way down here. You can simply get the User in the view `user = User.object.get(anonymous_user_id=<sub_id>)` and than pass user.id down into the existing functions we have. More code reuse and we don't have multiple concepts of what user_id means.

The extra DB call is pretty much irrelevant performance wise. We don't need to over-optimize such that everything is one ORM call.

ilee2u:
Small decisions note: the statuses in which an exam attempt can be flagged (ready_to_start, started, ready_to_submit, timed_out, submitted) are subject to change depending on the functionality that our proctoring vendors request in the future.

edx_exams/apps/core/api.py
```diff
@@ -48,6 +48,27 @@ def get_latest_attempt_for_user(user_id):
     return latest_attempt
 
 
+def get_user_by_anonymous_id(anonymous_user_id, attempt_number, resource_id):
```
Zacharis278, line 51: This function name seems to imply this returns the user for an anonymous id but it in fact returns an exam attempt. Overall comment on this is similar to my previous tho. Any way we can shoot for some code reuse here? We already have an api function that gives us an attempt for user, resource_id, and attempt_number which has near identical code.

If this just returned the user we could reuse that existing function based on the id of that user.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -155,9 +387,9 @@ def test_start_proctoring_unauthorized_user(self, mock_get_lti_launch_url):  # p
 
 @patch('edx_exams.apps.lti.views.get_lti_1p3_launch_start_url', return_value='https://www.example.com')
 class LtiEndAssessmentTestCase(ExamsAPITestCase):
-    """
+    '''
```
Zacharis278, line 390: Convention across our repos has docstrings with double quotes so I'd recommend we stick with that.

edx_exams/apps/lti/views.py
MichaelRoytman: You could, but how would you use it? You're checking the condition here, so this is where you'd raise the exception. But this is also the place you'd handle the exception.

I think this would be more useful if you were, say, performing a status transition in the API or something and needed to percolate up an exception to the calling code. We don't have a need for that yet, in my opinion.

edx_exams/apps/lti/views.py
```diff
@@ -12,19 +15,113 @@
...
 from edx_exams.apps.core.exceptions import ExamIllegalStatusTransition
+from edx_exams.apps.core.models import User
 from edx_exams.apps.core.statuses import ExamAttemptStatus
 from edx_exams.apps.lti.utils import get_lti_root
 
+log = logging.getLogger(__name__)
+
 EDX_OAUTH_BACKEND = 'auth_backends.backends.EdXOAuth2'
 
+LTI_PROCTORING_ASSESSMENT_CONTROL_ACTIONS = [
```
MichaelRoytman, line 40: You aren't using this constant. Do you still need it? Is it meant to be renamed `VALID_STATUSES`.

If so, keep in mind that you have some mismatches between the values in this list and what the code is looking for. For example, this variable has the value `flagRequest`, but the view code is checking for `flag`.

edx_exams/apps/lti/views.py
```diff
@@ -78,9 +175,16 @@ def start_proctoring(request, attempt_id):
         reverse('lti_consumer:lti_consumer.start_proctoring_assessment_endpoint')
     )
 
+    assessment_control_url = urljoin(
+        get_lti_root(),
+        reverse('lti:acs', kwargs={'lti_config_id': lti_config_id}),  # if the url actually existed
```
MichaelRoytman, line 180: Nit: I assume the comment is a note to self, so please remove it.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -1,20 +1,248 @@
...
+from lti_consumer.models import LtiConfiguration, LtiProctoringConsumer
 
 from edx_exams.apps.api.test_utils import ExamsAPITestCase, UserFactory
 from edx_exams.apps.core.models import CourseExamConfiguration, Exam, ExamAttempt
 from edx_exams.apps.core.statuses import ExamAttemptStatus
 from edx_exams.apps.lti.utils import get_lti_root
 
 
+@ddt.ddt
+class LtiAcsTestCase(ExamsAPITestCase):
```
MichaelRoytman, line 24: What do you think about refactoring this a little to use some of the [factories](https://github.com/edx/edx-exams/blob/main/edx_exams/apps/api/test_utils/factories.py) Zach introduced?

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -1,20 +1,248 @@
...
+                f'for lti config id {self.test_provider.lti_configuration_id}, exam id {self.exam.id}, and '
+                f'attempt id {self.attempt.id}. It has either not started yet, been rejcected, expired, or '
+                f'already verified.'
+            )
+
+        self.assertEqual(response.status_code, expected_response_status)
+        self.assertEqual(response.data, expected_msg)
+        mock_get_attempt.assert_called_once_with(self.user.id, self.attempt.attempt_number, self.exam.resource_id)
+
+    # TEST 3: get_attempt returned None -> Expected log
```
Zacharis278, line 197: If you want to describe what the test is doing you're better off using a function docstring to do this. The linter should make you do it anyway.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -1,20 +1,248 @@
...
+            'attempt_number': self.attempt.attempt_number,
+            'action': 'flag',
+            'incident_time': '2018-02-01T10:45:33Z',
+            'incident_severity': '0.1',
+            'reason_code': '12056',
+            'reason_msg': 'Excessive background noise outside candidate control'
+        }
+
+        # Even though the client.post function below uses json.dumps to serialize the request as json,
+        # The json serialization needs to happen before the request for an unkown reason
```
MichaelRoytman, line 172:
```suggestion
        # The json serialization needs to happen before the request for an unknown reason
```

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -1,20 +1,248 @@
...
+                f'with resource id {self.exam.resource_id} and attempt number {self.attempt.attempt_number} '
+                f'for lti config id {self.test_provider.lti_configuration_id}, exam id {self.exam.id}, '
+                f'and attempt id {self.attempt.id}.'
+            )
+        else:
+            expected_msg = (
+                f'Attempt cannot be flagged for user with anonymous id {self.user.anonymous_user_id} '
+                f'with resource id {self.exam.resource_id} and attempt number {self.attempt.attempt_number} '
+                f'for lti config id {self.test_provider.lti_configuration_id}, exam id {self.exam.id}, and '
+                f'attempt id {self.attempt.id}. It has either not started yet, been rejcected, expired, or '
```
MichaelRoytman, line 189: If you correct `rejcected` to `rejected`, this will fail, as a heads up.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -1,20 +1,248 @@
...
+                f'already verified.'
+            )
+
+        self.assertEqual(response.status_code, expected_response_status)
+        self.assertEqual(response.data, expected_msg)
+        mock_get_attempt.assert_called_once_with(self.user.id, self.attempt.attempt_number, self.exam.resource_id)
```
MichaelRoytman, line 195: I don't know that you need to test this. The most important thing to test is _what_ has happened, not _how_ it happened. I don't think it's important to know that this particular function was called. But please let me know if you disagree!

edx_exams/apps/lti/tests/test_views.py
MichaelRoytman, line 197: Nit: Please remove this comment.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -1,20 +1,248 @@
...
+            'attempt_number': false_attempt_number,
+            'action': 'flag',
+            'incident_time': '2018-02-01T10:45:33Z',
+            'incident_severity': '0.1',
+            'reason_code': '12056',
+            'reason_msg': 'Excessive background noise outside candidate control'
+        }
+
+        # Even though the client.post function below uses json.dumps to serialize the request as json,
+        # The json serialization needs to happen before the request for an unkown reason
```
MichaelRoytman, line 230:
```suggestion
        # The json serialization needs to happen before the request for an unknown reason
```

edx_exams/apps/lti/tests/test_views.py
Zacharis278, line 24: we could probably use some limited testing on auth here. It's mostly handled in the library but it's common to have something basic to make sure everything configured properly. For example when we do: https://github.com/edx/edx-exams/blob/e3e375993afe46b0bee6db91b163f5f0ee10a69b/edx_exams/apps/api/v1/tests/test_views.py#L82

Maybe in this case we just check for no token and a token without the correct scope.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -1,20 +1,248 @@
...
+            content_id=self.content_id,
+            exam_name='test_exam',
+            exam_type='proctored',
+            time_limit_mins=30,
+            due_date='2021-07-01 00:00:00',
+            hide_after_due=False,
+            is_active=True
+        )
+
+        self.attempt = ExamAttempt.objects.create(
```
Zacharis278, line 54: might want to take a look at the Factory classes added with #129. I think you can do the whole setup of this config/exam/attempt with only a couple lines.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -1,20 +1,248 @@
...
+                                  mock_authentication):  # pylint: disable=unused-argument
+        false_attempt_number = '88888888'
+
+        mock_get_attempt.return_value = None
+        mock_permissions.return_value = True
+
+        token = self._make_access_token('https://purl.imsglobal.org/spec/lti-ap/scope/control.all')
+
+        # Request w/ attempt number for an attempt that does not exist
+        request_body = {
```
Zacharis278, line 213: setting up this request body and making the request is the same across both of our tests here. Maybe a helper function to do this? Especially if we're adding another test for auth.

edx_exams/apps/lti/tests/test_views.py
ilee2u: I just pushed some code with modifications regarding all of the other comments. I also created a test for when authentication fails.

I tried making a test where the token is valid but doesn't have the correct scope and had some issues (I can go into more detail about this if you'd like). I noticed that we already test the permissions class for the correct scope within xblock-lti-consumer ([Link](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/tests/extensions/rest_framework/test_permissions.py#L281)), so I'm wondering if there's a reason we should still test for the scope within this repo.

ilee2u:
Notes for latest push:

- Streamlined tests to only test for higher level view functionality
- Added a basic authentication failure test
- Implemented the new factories that were added by Zach in a previous PR into my tests
- Other various cleanup items

edx_exams/apps/lti/tests/test_views.py
Zacharis278: ahh ok usually it's pretty easy to test so we do but I see how you'd have to setup an entire config with a non-proctoring consumer which is weird so let's leave this one out.

Zacharis278 (approved):
looking good! 🚀

PR #128, opened 2023-05-09
feat: update latest attempt endpoint

varshamenon4:
**JIRA:** [MST-1859](https://2u-internal.atlassian.net/browse/MST-1859)

**Description:** This PR updates the get latest endpoint to return latest regardless of status and to fix logic so that edx-proctoring call return of no active attempts doesn't return logic to return regardless of status.

**Testing instructions:**

1. [Set up LTI](https://2u-internal.atlassian.net/wiki/spaces/PT/pages/256737327/Local+Development+and+LTI+Configuration)
2. Open proctored exam as learner
3. Click through interstitials and click on Start System Check
4. Once LTI check passes, click Start Exam and should go to next ready to start interstitial
5. If reload is required before showing next interstitial, then failed check

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**


edx_exams/apps/api/v1/views.py
```diff
@@ -426,19 +426,19 @@ def get(self, request):
...
-            serialized_attempt = StudentAttemptSerializer(latest_attempt)
-
             if latest_attempt.status in (ExamAttemptStatus.started, ExamAttemptStatus.ready_to_submit):
                 return Response(status=status.HTTP_200_OK, data=serialized_attempt.data)
 
         # if edx-proctoring has an active attempt, return it
         latest_attempt_legacy, response_status = get_active_exam_attempt(user.lms_user_id)
         if latest_attempt_legacy is not None and response_status == status.HTTP_200_OK:
-            return Response(status=status.HTTP_200_OK, data=latest_attempt_legacy)
+            if latest_attempt_legacy != {}:
```
varshamenon4, line 440: Note: currently, when making call to edx-proctoring, {} is returned if there is no active attempt. This if statement, ensures that the legacy attempt is only returned if there is an active attempt.

varshamenon4 (commented):
Need to update tests. But this is working locally.

edx_exams/apps/core/tests/test_api.py
```diff
@@ -441,14 +441,6 @@ def test_no_attempt_for_user(self):
 
         self.assertIsNone(get_latest_attempt_for_user(9999999999))
 
-    def test_no_attempts_have_start_time(self):
```
varshamenon4, line 444: Removed this test since we now return attempts if they have no start time.

edx_exams/apps/core/tests/test_api.py
varshamenon4, line 1: Are there other tests/corner cases I should account for in the unit testing?

edx_exams/apps/api/v1/views.py
```diff
@@ -426,19 +426,19 @@ def get(self, request):
...
         if latest_attempt is not None:
             latest_attempt = check_if_exam_timed_out(latest_attempt)
-            serialized_attempt = StudentAttemptSerializer(latest_attempt)
-
             if latest_attempt.status in (ExamAttemptStatus.started, ExamAttemptStatus.ready_to_submit):
                 return Response(status=status.HTTP_200_OK, data=serialized_attempt.data)
 
         # if edx-proctoring has an active attempt, return it
         latest_attempt_legacy, response_status = get_active_exam_attempt(user.lms_user_id)
         if latest_attempt_legacy is not None and response_status == status.HTTP_200_OK:
```
Zacharis278, line 439: looks like this could just be merged with the new condition so we only have a single if statement here. That would make the line long but you could refactor to this format:
```suggestion
        if (
            latest_attempt_legacy is not None
            and latest_attempt_legacy != {}
            and response_status == status.HTTP_200_OK
        ):
```

Zacharis278 (changes requested):
Once small suggestion otherwise looks pretty good.

I think there's still a gap here but it may be 'unsolvable' in this API alone. If we have multiple exam attempts in a created or download clicked state the most recently created one may not be the exam the user is currently viewing. One possible solution I can think of quickly is to add an optional exam_id filter to this request the UI can use in certain cases. There might be other solutions tho.

This gap is more of an edge case that generally won't block testing the overall flow so I'll leave it to you if you want to try and tackle it now or open a bug ticket for later.

edx_exams/apps/core/tests/test_api.py
Zacharis278: we should probably have a test that covers the case where there are multiple attempts but none of them have a start time. So really similar to `test_get_latest_exam_attempt_for_user`

edx_exams/apps/api/v1/views.py
Zacharis278: good find!

edx_exams/apps/api/v1/views.py
```diff
@@ -426,18 +426,22 @@ def get(self, request):
         """
         user = request.user
         latest_attempt = get_latest_attempt_for_user(user.id)
+        serialized_attempt = StudentAttemptSerializer(latest_attempt)
 
         # if there is an active attempt in this service, return it
         if latest_attempt is not None:
             latest_attempt = check_if_exam_timed_out(latest_attempt)
             serialized_attempt = StudentAttemptSerializer(latest_attempt)
```
varshamenon4, line 434: Note: I'm keeping the serialized attempt, since the line above (check_if_exam_timed_out) can potentially update the attempt status in which case the serialized attempt will be different from the originally set one.

varshamenon4 (commented):
Added tests and made changes based on feedback.

varshamenon4:
> Once small suggestion otherwise looks pretty good.
> 
> I think there's still a gap here but it may be 'unsolvable' in this API alone. If we have multiple exam attempts in a created or download clicked state the most recently created one may not be the exam the user is currently viewing. One possible solution I can think of quickly is to add an optional exam_id filter to this request the UI can use in certain cases. There might be other solutions tho.
> 
> This gap is more of an edge case that generally won't block testing the overall flow so I'll leave it to you if you want to try and tackle it now or open a bug ticket for later.

I see. I've gone ahead and created a bug ticket for this to tackle separately: https://2u-internal.atlassian.net/browse/MST-1901

alangsto (approved):
nice 👍

PR #141, opened 2023-06-08
feat: add context to launch data

varshamenon4:
**JIRA:** MST-1932

**Description:** This PR adds context (content url and MFE url as context_label and context_title respectively) to the launch data so that it can be used to redirect once an LTI proctored exam is started.

**Dependencies:** Blocked by https://github.com/openedx/xblock-lti-consumer/pull/382

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/lti/views.py
+++ b/edx_exams/apps/lti/views.py
@@ -183,6 +183,7 @@ def start_proctoring(request, attempt_id):
         message_type='LtiStartProctoring',
         proctoring_launch_data=proctoring_launch_data,
         context_id=exam.course_id,
+        context_label=exam.content_id,
     )
 
     return redirect(get_lti_1p3_launch_start_url(launch_data))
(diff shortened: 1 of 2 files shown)


edx_exams/apps/lti/views.py
```diff
@@ -183,6 +184,8 @@ def start_proctoring(request, attempt_id):
         message_type='LtiStartProctoring',
         proctoring_launch_data=proctoring_launch_data,
         context_id=exam.course_id,
+        context_title=settings.LEARNING_MICROFRONTEND_URL,
```
varshamenon4, line 187: Question: I'm using context title to pass along the MFE url but let me know if there is a better way to pass along in launch data so that it can be used in xblock lti consumer.

alangsto: If it's already in settings, you don't need to pass it to xblock-lti-consumer. That plugin will also have access to the same settings

PR #146, opened 2023-06-27
feat: launch instructor tool

Zacharis278:
### [MST-1948](https://2u-internal.atlassian.net/browse/MST-1948)

Creates a new view that will initiate a basic LTI 1.3 launch into the proctoring vendor's instructor tool. I opted to make this a view since returning the URL just adds an extra step to the UI to fetch it and stick it in an iframe. This way we just load up an iframe with `src=http://edx-exams.edx.org/lti/exam/28/instructor_tool` and this view does the rest.

Original concept was to build a generic launch view into the lti-consumer library based on this POC https://github.com/openedx/xblock-lti-consumer/pull/386. That PR documents the issues I found with that approach.

### DIFF
--- a/edx_exams/apps/lti/tests/test_views.py
+++ b/edx_exams/apps/lti/tests/test_views.py
@@ -17,7 +17,10 @@ from lti_consumer.models import LtiConfiguration, LtiProctoringConsumer
 
 from edx_exams.apps.api.test_utils import ExamsAPITestCase, UserFactory
 from edx_exams.apps.api.test_utils.factories import (
-    CourseExamConfigurationFactory, ExamAttemptFactory, ExamFactory, ProctoringProviderFactory
+    CourseExamConfigurationFactory,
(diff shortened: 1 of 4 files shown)


edx_exams/apps/lti/tests/test_views.py
```diff
@@ -211,32 +216,14 @@ def setUp(self):
...
 
-        self.course_exam_config = CourseExamConfiguration.objects.create(
-            course_id=self.course_id,
-            provider=self.test_provider,
-            allow_opt_out=False
-        )
-
-        self.exam = Exam.objects.create(
-            resource_id=str(uuid.uuid4()),
+        self.exam = ExamFactory(
```
Zacharis278, line 219: this is just a refactor to remove more verbose data setup and replace it with factory methods.

alangsto (approved):
One nit, otherwise LGTM 👍

PR #154, opened 2023-07-19
Add "terminate" action handling to ACS endpoint

ilee2u:
**JIRA:** https://2u-internal.atlassian.net/jira/software/projects/MST/boards/584?selectedIssue=MST-1961
Description

As of now, the ACS endpoint only supports a flag action. We would also like to support terminate as a way for Proctorio to communicate that they have ended an attempt. 

Upon receiving a terminate request, the attempt referenced in the request should be moved to a corresponding status, depending on the reason for termination (reason_code) and incident_severity. The reason_code mappings should be clarified with Proctorio, and the incident_severity should be discussed internally to decide what range of values are considered passing. 

Reason codes:

user_submission: move attempt to verified or second_review_required

error: move attempt to error

Severity:

we can just use a constant set to 0.25 for this at the moment

terminate with severity > 0.25 will move to second_review_required otherwise verified

Question: Do we want to store the reason code and incident severity in another model (similar to a review)?

work with dev on MST-1992: Return review info to exams-dashboard
IN REVIEW
 

**Description:** Describe in a couple of sentences what this PR adds

migrations, etc.

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**


edx_exams/apps/core/statuses.py
```diff
@@ -44,6 +44,9 @@ class ExamAttemptStatus:
     # the course end date has passed
     expired = 'expired'
 
+    # the exam attempt requires a secondary review to verify its integrity
```
Zacharis278, line 47: think this is a duplicate of the existing one on line 39

edx_exams/apps/lti/views.py
```diff
@@ -101,14 +101,111 @@ def acs(request, lti_config_id):
...
+    elif action == 'terminate':
+        # Upon receiving a terminate request, the attempt referenced should have their status updated
+        # depending on the reason for termination (reason_code) and the incident_severity (scaling from 0.0 to 1.0).
+        # If the severity is greater than 0.25, then the attempt is marked for secondary review.
+
+        # Get the reason code for the termination and ensure it's a string to comply with the LTI specs
+        # See: http://www.imsglobal.org/spec/proctoring/v1p0#h.rsq8h6qxveab
+        reason_code_key = data['reason_code']
+        if not isinstance(reason_code_key, str):
+            error_msg = 'ERROR: reason_code must be passed to the ACS endpoint as a string per LTI specs.'
```
Zacharis278, line 124: might want to just change this to 'reason_code is required'. It's optional to implement by the spec but edx does require it.

edx_exams/apps/lti/views.py
```diff
@@ -101,14 +101,111 @@ def acs(request, lti_config_id):
...
+        if not isinstance(severity, str):
+            error_msg = 'ERROR: incident_severity must be passed to the ACS endpoint as a string per LTI specs.'
+            error = {'detail': error_msg}
+            return Response(status=status.HTTP_400_BAD_REQUEST, data=error)
+
+        # Convert reason code and severity to correct formats
+        # NOTE: Is there a point in checking if severity is a string if we're just converting it into a
+        # float anyways?
+        reason_code = ASSESSMENT_CONTROL_CODES[reason_code_key]
+        severity = float(severity)
```
Zacharis278, line 139: I think a decimal may be safer to use here when parsing from a string.

edx_exams/apps/lti/views.py
```diff
@@ -101,14 +101,111 @@ def acs(request, lti_config_id):
...
+        SEVERITY_THRESHOLD = 0.25
+        success_msg = ''
+
+        # Reason codes:
+        # Regular submission occurred
+        # Code 1
+        if reason_code == 'Submitted by user':
```
Zacharis278, line 146: I think we can simplify this down to '1' means success and anything else is an error. The strings in that list are basically UI messaging so they could change and break this function. In any case checking against "magic strings" is usually not a great idea.

I don't think there's any distinction in behavior for the various errors on our end, it's mostly just information for the instructor. A non '1' response should always result in the error state regardless of severity. The instructor can figure out if it's a pass/fail/retake from there.

edx_exams/apps/lti/views.py
```diff
@@ -101,14 +101,111 @@ def acs(request, lti_config_id):
...
+            update_attempt_status(attempt.id, ExamAttemptStatus.verified)
```
Zacharis278, line 147: looks like we aren't checking severity on the successfully submitted case

edx_exams/apps/lti/views.py
```diff
@@ -172,7 +269,8 @@ def start_proctoring(request, attempt_id):
         attempt_number=attempt.attempt_number,
         start_assessment_url=proctoring_start_assessment_url,
         assessment_control_url=assessment_control_url,
-        assessment_control_actions=['flagRequest'],
+        # TODO: Is this the correct action?
+        assessment_control_actions=['terminateRequest'],
```
Zacharis278, line 273: it's actually just `terminate` the LTI docs are missing a space making it seems like 'Request' is part of the string.

edx_exams/apps/lti/views.py
Zacharis278: side note maybe we can request they swap the 0 and 1 error codes. Would make a lot more sense since that's already a convention in exit codes.

edx_exams/apps/lti/views.py
```diff
@@ -101,14 +103,98 @@ def acs(request, lti_config_id):
...
+
+        # Ensure the incident_severity's a string to comply with the LTI specs
+        # See: http://www.imsglobal.org/spec/proctoring/v1p0#h.rsq8h6qxveab
+        if not isinstance(severity, str):
+            error_msg = 'ERROR: incident_severity must be passed to the ACS endpoint as a string per LTI specs.'
+            return Response(status=status.HTTP_400_BAD_REQUEST, data=error_msg)
+
+        severity = Decimal(severity)
+        SEVERITY_THRESHOLD = 0.25
+        reason_code_description = ASSESSMENT_CONTROL_CODES[reason_code]
```
Zacharis278, line 141: If this is only feeding log messages I don't think we need to import these. Two reasons, they are a bit verbose for log messages when the reason_code will do and importing from api is a bit of an odd pattern of app dependencies (if these are used everywhere the strings should move to core)

edx_exams/apps/lti/views.py
```diff
@@ -101,14 +103,98 @@ def acs(request, lti_config_id):
...
+        # Regular submission occurred, but the learner did something
+        # that might be worth marking the attempt for review. Mark the attempt
+        # as requiring review based on the severity level (>0.25 -> needs review)
+        if reason_code == '1':
+            if severity > SEVERITY_THRESHOLD:
+                update_attempt_status(attempt.id, ExamAttemptStatus.second_review_required)
+                success_msg = (
```
Zacharis278, line 148: missing a log statement for this message. But, maybe it would be better to just call log.info() once at the end of these cases?

Also seems like most of these log messages have similar content could we maybe just log the repetitive part on it's own once? `Terminating exam attempt for user ... attempt id {attempt.id}.`

I see the whole log string is sent back to the vendor in the response, I don't think we need to send all of that. The LTI docs have what appears to be optional status field to send back so if we send anything it should be a JSON object just so that response type doesn't change if we use that later.

edx_exams/apps/lti/views.py
```diff
@@ -101,14 +103,98 @@ def acs(request, lti_config_id):
...
+                    f'Termination Severity > 0.25, marking exam attempt for secondary review. '
+                    f'Terminating exam attempt for user with id {anonymous_user_id} '
```
Zacharis278, line 150: I'd make this log clear this is `anonymous_user_id ` and not just user_id

edx_exams/apps/lti/views.py
```diff
@@ -101,14 +103,98 @@ def acs(request, lti_config_id):
...
+        # Get the termination paramenters
+        try:
+            reason_code = data['reason_code']
+            incident_time = data['incident_time']
+            severity = data['incident_severity']
+        except KeyError as err:
+            error_msg = f'ERROR: required parameter {err} was not found.'
+            return Response(status=status.HTTP_400_BAD_REQUEST, data=error_msg)
+
+        # Ensure the incident_severity's a string to comply with the LTI specs
```
Zacharis278, line 133: tbh if they send a Number we can still parse it so I don't see much benefit in being strict here. Just more code for us to maintain

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -193,6 +205,171 @@ def test_acs_no_attempt_found(self,
...
+
+    @ ddt.data(
+        # Testing reason codes with severity > 0.25
+        ('0', '1.0', 'error'),
+        ('1', '1.0', 'second_review_required'),
+        ('2', '1.0', 'error'),
+        ('4', '1.0', 'error'),
+        ('5', '1.0', 'error'),
+        ('6', '1.0', 'error'),
+        ('7', '1.0', 'error'),
```
Zacharis278, line 309: this is going to produce a lot of tests which do take time to run. I don't think testing every value has much benefit

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -38,7 +39,11 @@ def setUp(self):
 
         self.course_exam_config = CourseExamConfigurationFactory()
         self.exam = ExamFactory()
+
+        # Create an Exam Attempt that has already been submitted.
         self.attempt = ExamAttemptFactory()
+        self.attempt.status = ExamAttemptStatus.submitted
```
Zacharis278, line 45: you can pass fields into ExamAttemptFactory to so it's created with status=ExamAttemptStatus.submitted instead of the default.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -193,6 +205,171 @@ def test_acs_no_attempt_found(self,
...
+        token = self.make_access_token('https://purl.imsglobal.org/spec/lti-ap/scope/control.all')
+
+        request_body = self.create_request_body(
+            self.attempt.attempt_number,
+            'terminate',
+            reason_code,
+            incident_severity
+        )
+
+        # Even though the client.post function below uses json.dumps to serialize the request as json,
```
Zacharis278, line 360: looks like we're repeating this same block to do this post in multiple places. My rule of thumb, if you repeat it 3 or more times consider a helper function.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -193,6 +205,171 @@ def test_acs_no_attempt_found(self,
...
+        # The json serialization needs to happen before the request for an unknown reason
+        request_body = json.dumps(request_body)
+        self.client.post(self.url, data=request_body, content_type='application/json',
+                         HTTP_AUTHORIZATION='Bearer {}'.format(token))
+        self.attempt.refresh_from_db()
+
+        self.assertEqual(self.attempt.status, expected_attempt_status)
+
+        # Assure an entry was added to the ACResult model
+        data = AssessmentControlResult.objects.all()
```
Zacharis278, line 370: since we know the attempt id already an improvement to this test would to be to query for that result specifically.

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -179,20 +190,141 @@ def test_acs_no_attempt_found(self,
...
+                           incident_severity,
+                           expected_attempt_status,
+                           mock_get_attempt,
+                           mock_permissions,
+                           mock_authentication):  # pylint: disable=unused-argument
+        """
+        Test that the terminate action changes the exam attempt status as expected
+        based on the 'reason_code' and 'incident_severity'.
+        """
+        self.attempt.status = ExamAttemptStatus.submitted
```
Zacharis278, line 307: NIT: the test attempt is already submitted

edx_exams/apps/lti/tests/test_views.py
```diff
@@ -179,20 +190,141 @@ def test_acs_no_attempt_found(self,
...
+        else:
+            del request_body[acs_parameter][acs_sub_parameter]
+            key_to_fail = acs_sub_parameter
+
+        response = self.make_post_request(request_body, self.token)
+
+        self.attempt.refresh_from_db()
+        self.assertEqual(response.data, f'ERROR: required parameter \'{key_to_fail}\' was not found.')
+
+    # TODO: Add tests for termination errors
```
Zacharis278, line 244: NIT: old todo comment

Zacharis278 (approved):
Two nits but otherwise looking great 🚀

edx_exams/apps/lti/views.py
```diff
@@ -101,14 +102,80 @@ def acs(request, lti_config_id):
...
+        severity = Decimal(severity)
+        SEVERITY_THRESHOLD = 0.25
+        # Regular submission occurred, but the learner did something
+        # that might be worth marking the attempt for review. Mark the attempt
+        # as requiring review based on the severity level (>0.25 -> needs review)
+        if reason_code == '1':
+            if severity > SEVERITY_THRESHOLD:
+                update_attempt_status(attempt.id, ExamAttemptStatus.second_review_required)
+                success_msg = 'Termination Severity > 0.25, marking exam attempt for secondary review. '
+            elif severity <= SEVERITY_THRESHOLD:
```
Zacharis278, line 141: oh I think the way this if/elif is causing a coverage issue because there's no base case here. Maybe just stick to if/else?

PR #155, opened 2023-07-24
feat: verify/reject actions on attempt update

zacharis278:
**JIRA:** [MST-1988](https://2u-internal.atlassian.net/browse/MST-1988)

Allow staff users to perform the update actions of 'verify' and 'reject'


alangsto (approved):
Nice solution 👍

PR #158, opened 2023-08-03
feat: latest attempt endpoint fixes

Zacharis278:
**JIRA:** [MST-1950](https://2u-internal.atlassian.net/browse/MST-1950?atlOrigin=eyJpIjoiMmFlZTk4MjY2MjFiNDEwN2EzZWY5YWIxMzdkMjNmYWQiLCJwIjoiaiJ9)

Frontend PR: https://github.com/openedx/frontend-lib-special-exams/pull/113

Alter the behavior of this endpoint such that:

- By default it only returns an actively running attempt (the timer is counting). This means started or ready_to_submit. There should only ever be one such exam. This is what the MFE requests in almost every case.
- If a specific exam is requested, return the user's attempt for that attempt regardless of state. The UI only does this when performing checks on the download_click / start buttons to ensure the backend is in the expected state.

The story calls for using exam_id here but that could mislead devs troubleshooting or looking at logs in the future since an exam id could be duplicated between edx-exams and edx-proctoring. I chose content id instead since it's unique.

Hindsight, this probably would be less messy as two endpoints but I didn't want to go back on the decisions we've already made or rework the UI. Once edx-proctoring goes away so does 90% of this logic. I've left a note in code and opened https://2u-internal.atlassian.net/browse/MST-2047


edx_exams/apps/lti/signals/tests/test_handlers.py
```diff
@@ -8,9 +8,9 @@
 from ddt import ddt, idata, unpack
 from django.test import TestCase
 
-from edx_exams.apps.api.test_utils.factories import UserFactory
 from edx_exams.apps.core.models import Exam, ExamAttempt, ProctoringProvider
 from edx_exams.apps.core.statuses import ExamAttemptStatus
+from edx_exams.apps.core.test_utils.factories import UserFactory
```
Zacharis278, line 13: It made more sense for this util to live in core since we really wouldn't want the lti/core apps to depend on the api app.

edx_exams/apps/core/api.py
```diff
@@ -38,14 +38,11 @@ def get_attempt_by_id(attempt_id):
     return attempt
 
 
-def get_latest_attempt_for_user(user_id):
+def get_active_attempt_for_user(user_id):
```
Zacharis278, line 41: renamed this since it's behavior isn't really latest anymore

edx_exams/apps/api/v1/views.py
```diff
@@ -440,36 +444,50 @@ def get(self, request):
...
+        exam_content_id = request.GET.get('content_id', None)
+        exam = get_exam_by_content_id(exam_content_id)
+
+        # if a specific exam is requested always return that exam
+        if exam is not None:
+            attempt = get_current_exam_attempt(user.id, exam.id)
+        else:
+            attempt = get_active_attempt_for_user(user.id)
+
+        serialized_attempt = StudentAttemptSerializer(attempt)
```
ilee2u, line 467: Would it be more efficient for this to be below the if statement block below? I'm thinking we might otherwise end up serializing this attempt twice for no reason.

edx_exams/apps/api/v1/views.py
```diff
@@ -440,36 +444,50 @@ def get(self, request):
...
         ):
-            return Response(status=status.HTTP_200_OK, data=latest_attempt_legacy)
+            return Response(status=status.HTTP_200_OK, data=legacy_attempt)
 
-        # otherwise return the latest attempt here regardless of status
+        # otherwise return the attempt from edx-exams regardless of status
         return Response(
             status=status.HTTP_200_OK,
-            data=serialized_attempt.data if latest_attempt is not None else {}
+            data=serialized_attempt.data if attempt is not None else {}
```
ilee2u, line 490: Would it also be worth maybe checking if attempt is not None here, serializing the data, then sending it back in the response?

edx_exams/apps/api/v1/views.py
Zacharis278: good call, there's something odd about that logic here

ilee2u (approved):

PR #167, opened 2023-09-07
docs: ADR for downstream events

ilee2u:
**JIRA:** [MST-2082](https://2u-internal.atlassian.net/browse/MST-2082)

**Description:** Create an ADR for this M6 with VERY specific definitions of what ALL of these events we will produce/emit from edx-exams, including:

Why we choose to use event bus

The context under which they are emitted from edx-exams.

The event properties (i.e. data sent)

Where these events are received.

What effects/functions are triggered when similar events are received.

 

Use the M6 Spec Doc as a reference: [MST-1789: M6 (Exam Review and Downstream Triggers) Scope Definition and Story Writing | Event Definitions:](https://2u-internal.atlassian.net/wiki/spaces/PT/pages/539066520/MST-1789+M6+Exam+Review+and+Downstream+Triggers+Scope+Definition+and+Story+Writing#Event-Definitions%3A)

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/docs/decisions/0004-downstream-effect-events.rst
+++ b/docs/decisions/0004-downstream-effect-events.rst
@@ -20,8 +20,8 @@ As such, we plan to use the event bus to send info to edx-platform services with
 
 Below are lists of downstream effects of exam submission and review that we will or will not be translating into events in edx-exams as part of this decision.
 
-Downstream effects to be implemented as events in the short term:
-*****************************************************************
+Downstream effects to be implemented as events:
+***********************************************
 
   * Grades Override - A python API call to the grades app to generate a grade override when an exam attempt is rejected.
 
@@ -35,8 +35,8 @@ Downstream effects to be implemented as events in the short term:
 
   * Set Credit Requirement Status - A python API call to the credits app to create or modify a learner’s credit requirement status for an exam.
 
-Downstream effects that will be implemented in the long term:
-*************************************************************
+Downstream effects that are not being implemented as part of this decision:
+***************************************************************************
 
   * Credit Prerequisite Check:
 


docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+#. Event consumers added to edx-platform
+
+  * We will add consumers in the signals.py file in each edx-platform service's respective folders.
+
+  * These consumers will call other service or api functions in those folders.
+
+References
+----------
+
+* Discovery Doc for M6: https://2u-internal.atlassian.net/wiki/spaces/PT/pages/539066520/MST-1789+M6+Exam+Review+and+Downstream+Triggers+Scope+Definition+and+Story+Writing#Implementation:
```
MichaelRoytman, line 101: This document is not going to be visible to the Open edX community because it's in 2U's internal Confluence instance. We should either not link to it or consider moving this to the Open edX Confluence space.

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+=======================================================================
+
+Status
+------
+
+Draft (circa Sept 2023)
+
+Context
+-------
+For Milestone 6 of the exams IDA project, we are porting any downstream effects of exam submission and review
```
MichaelRoytman, line 11: Nit. Since the audience for this ADR is 2U and Open edX, I don't think we need to make reference to internal concepts like milestones here. It may be confusing, and it's not relevant to the actual subject of the ADR.

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+from edx-proctoring to the exams IDA (aka edx-exams). Most of these downstream effects will reflect calls
+that edx-proctoring (which is a plugin) makes to various edx-platform services, such as grades, credits, and certificates.
+
+Since edx-exams is independent, we plan to use the event bus to send info to edx-platform services
```
schenedx, line 15: This is a decision record, let's use more strong language:
"we decided to use the event bus to...."

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+without needing a response as one would in a REST setup.
+
+This is being done to circumvent messy circular dependencies, wherein edx-exams and edx-platform go back and
+forth asking each other for information.
+
+
+Decision
+--------
+We are porting these because they are both essential and easy to translate into events:
```
MichaelRoytman, line 24: Can you be more specific about what "these" refers to? "These" refers to "downstream effects of exam submission and review", right? It would be useful to tie those two paragraphs together.

docs/decisions/0004-downstream-effect-events.rst
schenedx, line 11: You are referencing the plans and name of the project, but you did not reference a project overview/summary. Imagine the confusion for others outside of Cosmonauts team.
Either don't reference the project, or provide a link to the project overview document

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+
+    * Instead, we have decided to implement an endpoint in the credits service that returns the prereq status, which will be called directly from the exams UI.
+
+    * This feature will be developed separately and is considered to be outside of the scope of M6.
+
+  * Name Affirmation:
+
+    * Currently, edx-proctoring calls the name affirmation service in order to match the name that Proctortrack sees on the learner’s ID to the user’s verified name in the LMS database.
+
+    * We do not plan to implement this since It is very likely that 2U will not require the matching of one’s verified name in edx-platform to the names on their IDs in the future.
```
MichaelRoytman, line 56: Nit.

```suggestion
    * We do not plan to implement this since it is very likely that 2U will not require the matching of one’s verified name in edx-platform to the names on their IDs in the future.
```

docs/decisions/0004-downstream-effect-events.rst
schenedx: This is no longer the context. This should be in the "Decisions" paragraph. You should explain why edx-exam is independent

docs/decisions/0004-downstream-effect-events.rst
MichaelRoytman: What consideration has been made to the Open edX community. I don't think Name Affirmation has wide adoption outside of 2U, but is there any concern about making this change on Open edX operators? It could be worth making a post in the Open edX Discuss forum to inform them of this decision, unless you're already planning to circulate this ADR there.

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+Context
+-------
+For Milestone 6 of the exams IDA project, we are porting any downstream effects of exam submission and review
+from edx-proctoring to the exams IDA (aka edx-exams). Most of these downstream effects will reflect calls
+that edx-proctoring (which is a plugin) makes to various edx-platform services, such as grades, credits, and certificates.
+
+Since edx-exams is independent, we plan to use the event bus to send info to edx-platform services
+without needing a response as one would in a REST setup.
+
+This is being done to circumvent messy circular dependencies, wherein edx-exams and edx-platform go back and
```
schenedx, line 18: This paragraph is impact of the decision. This should be in the consequences section.

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Consequences
+------------
+#. Event definitions implemented in openedx-events
+
+  * Defining the events and the data sent in each in this abstraction layer is fundamental to making event bus work.
+
+  * We have designed these events to be "generic", such that they can be triggered under contexts outside of exams by other services.
+
+#. Event producers implemented to edx-exams
```
MichaelRoytman, line 88: Nit.

```suggestion
#. Event producers implemented in edx-exams
```

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+
+    * Furthermore, it will likely take a very significant amount of time for Verificient to implement ID Verification in LTI form, so we do not need to do this until we know that they are implementing that feature on their end.
+
+    * Ideally, we want to be able to call the name affirmation service to inject the user’s first and last name into the LTI launch data data as the “User’s Verified Name”, but the logic regarding verified names isn’t even coded out in xblock-lti-consumer because the value of such names are unclear right now.
+
+  * Enrollments:
+
+    * This is only called to get a learner’s onboarding status.
+
+    * We will NOT be implementing this, since we are not implementing onboarding exams to reach MVP.
```
schenedx, line 66: You will not be implementing these as part of this project. But someone else can come a long and implement. Therefore the language should be:
"The following are not implemented as part of this decision".

schenedx (changes requested):
Added comments. Thank you for documenting down this important step.

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Status
+------
+
+Draft (circa Sept 2023)
+
+Context
+-------
+For Milestone 6 of the exams IDA project, we are porting any downstream effects of exam submission and review
+from edx-proctoring to the exams IDA (aka edx-exams). Most of these downstream effects will reflect calls
```
MichaelRoytman, line 12: I generally like to steer away from using and acronym until it has been defined because not everyone is familiar with what acronyms represent.  Could you define IDA first before using it? You could even just use "service" instead.

MichaelRoytman (changes requested):
Thanks for putting this together! These are just some thoughts on a first pass. I'll take another pass when I can find some more time.

docs/decisions/0004-downstream-effect-events.rst
ilee2u: I'm not sure if name affirmation for learner IDs has any impact on Open edX operators. What would be the best way to find out if changing this would be a concern?

Zacharis278: I'd leave 2U intentions out of this entirely. We're mostly leaving this out to keep initial scope light but I wouldn't call it a decision to drop name affirmation support. We can just say it is left out defining this one as we don't believe it to be an essential feature.

docs/decisions/0004-downstream-effect-events.rst
ilee2u: I'm a bit confused by this, as I do see some links to the internal confluence in the other ADR files within this folder, e.g. https://github.com/edx/edx-exams/blob/main/docs/decisions/0002-multiple-exam-systems.rst?plain=1#L54

schenedx: Yeah, I think we have been inconsistent with this. I don't think the feedback here is based on opinions you don't have to adopt. You can do whatever you feel is easy.

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+
+  * Instructor Complete Attempt - A call to the instructor service that marks the exam as completed.
+
+  * Instructor Is Course Staff - A call to the instructor service to check if a user is course staff.
+
+  * Invalidate Certificate - A call to the certificates service to invalidate a leaner’s edx certificate for a course.
+
+  * Set Credit Requirement Status - A call to the credits service to create or modify a learner’s credit requirement status for an exam.
+
+Here’s what we’re NOT porting (and why):
```
Zacharis278, line 40: An ADR isn't always a commitment to completely finish the development it just sets a direction. So things that might happen and in different project or milestone can still be consequences of the decision.

I think the language here should define what should be built instead of stating what 'we' are specifically committing to do.

Instead of "we have decided to implement X" say "x will be implemented"

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+
+    * This feature will be developed separately and is considered to be outside of the scope of M6.
+
+  * Name Affirmation:
+
+    * Currently, edx-proctoring calls the name affirmation service in order to match the name that Proctortrack sees on the learner’s ID to the user’s verified name in the LMS database.
+
+    * We do not plan to implement this since It is very likely that 2U will not require the matching of one’s verified name in edx-platform to the names on their IDs in the future.
+
+    * Furthermore, it will likely take a very significant amount of time for Verificient to implement ID Verification in LTI form, so we do not need to do this until we know that they are implementing that feature on their end.
```
Zacharis278, line 58: I'd leave this out entirely since it's a matter of projects/timelines and not architecture.

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,103 @@
...
+
+This is being done to circumvent messy circular dependencies, wherein edx-exams and edx-platform go back and
+forth asking each other for information.
+
+
+Decision
+--------
+We are porting these because they are both essential and easy to translate into events:
+
+  * Grades Override - A call to the grades service to override a grade when an exam attempt is rejected.
```
Zacharis278, line 26: I might suggest changing the language here a bit 'X Service' implies an independent deployment to me. This really calls a specific app within the LMS.  Maybe refer to the app name that 'service' class exists in? "A python api call to the XYZ app inside edx-platform"

docs/decisions/0004-downstream-effect-events.rst
Zacharis278, line 12: I think we can be more specific in what 'calls' to platform services means. The reader might falsely assume this is a network request.

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -0,0 +1,105 @@
...
+
+  * Instructor Delete Attempt - A python API call to the instructor app to delete an exam attempt.
+
+  * Instructor Complete Attempt - A python API call to the instructor app to mark an exam subsection as completed.
+
+  * Invalidate Certificate - A python API call to the certificates app to invalidate a leaner’s edx certificate for a course.
+
+  * Set Credit Requirement Status - A python API call to the credits app to create or modify a learner’s credit requirement status for an exam.
+
+Downstream effects that will be implemented in the long term:
```
schenedx, line 38: You should be explicit that "in the long term" means, they are not implemented as part of accepting this decision. Things changes in the future. I just don't want the record read as a commitment to follow up our long term plans.

schenedx (approved):
👍

Zacharis278 (approved):
This looks great! 👍 

I think you can move the status to accepted on merge since we've decided to move forward with this approach already.

MichaelRoytman (approved):
This looks really good - easy to read and understand and concise!

PR #171, opened 2023-09-12
feat: exam end return url

Zacharis278:
**JIRA:** [MST-2099](https://2u-internal.atlassian.net/browse/MST-2099)

**Description:** Sets a return url for the proctoring tool when performing an LtiEndAssessment launch

Required so tool can properly redirect learner back after ending the assessment once https://github.com/openedx/frontend-lib-special-exams/pull/117 is implemented.

### DIFF
--- a/edx_exams/apps/lti/views.py
+++ b/edx_exams/apps/lti/views.py
@@ -8,7 +8,6 @@ from decimal import Decimal
 from urllib.parse import urljoin
 
 from django.contrib.auth import login
-from django.http import JsonResponse
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.views.decorators.http import require_http_methods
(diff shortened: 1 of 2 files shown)


edx_exams/apps/lti/views.py
```diff
@@ -303,6 +298,11 @@ def end_assessment(request, attempt_id):
     # Platform MUST send an End Assessment message to the Proctoring Tool. Otherwise, the Assessment Platform can
     # complete its normal post-assessment flow.
     if end_assessment_return:
+        # user is authenticated via JWT so use that to create a
```
Zacharis278, line 301: I moved this since it's only necessary if we're performing an end assessment launch

edx_exams/apps/lti/views.py
```diff
@@ -328,7 +329,7 @@ def end_assessment(request, attempt_id):
 
         return redirect(preflight_url)
 
-    return JsonResponse({})
+    return redirect(get_exam_url_path(exam.course_id, exam.content_id))
```
Zacharis278, line 332: this will leave us on a blank page with '{}'. If we don't need to perform an LtiEndAssessment launch return to the exam section.

alangsto (approved):

PR #178, opened 2023-09-20
Emit the ExamAttemptSubmitted Open edX event when an exam is submitted.

MichaelRoytman:
**JIRA:** [MST-2114](https://2u-internal.atlassian.net/browse/MST-2114) (private)

**Description:** This pull request introduces a change to emit the `TODO` Open edX event when a learner submits their exam.

**Author concerns:**  None.

**Dependencies:**

* https://github.com/openedx/openedx-events/pull/268

**Testing instructions:**

Follow the instructions in [How to start using the Event Bus](https://openedx.atlassian.net/wiki/spaces/AC/pages/3508699151/How+to+start+using+the+Event+Bus) to test out this event.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/core/api.py
+++ b/edx_exams/apps/core/api.py
@@ -17,6 +17,7 @@ from edx_exams.apps.core.exceptions import (
     ExamIllegalStatusTransition
 )
 from edx_exams.apps.core.models import Exam, ExamAttempt
+from edx_exams.apps.core.signals.signals import emit_exam_attempt_submitted_event
 from edx_exams.apps.core.statuses import ExamAttemptStatus
 
 log = logging.getLogger(__name__)
@@ -95,6 +96,15 @@ def update_attempt_status(attempt_id, to_status):
 
     if to_status == ExamAttemptStatus.submitted:
         attempt_obj.end_time = datetime.now(pytz.UTC)
+
+        course_key = CourseKey.from_string(attempt_obj.exam.course_id)
+        usage_key = UsageKey.from_string(attempt_obj.exam.content_id)
+        emit_exam_attempt_submitted_event(
+            attempt_obj.user,
+            course_key,
+            usage_key,
+            attempt_obj.exam.exam_type
+        )
 
     attempt_obj.status = to_status
     attempt_obj.save()
--- a/edx_exams/apps/core/signals/signals.py
+++ b/edx_exams/apps/core/signals/signals.py
@@ -1,0 +1,32 @@
+"""
+Signal definitions and functions to send those signals for the edx-exams application.
+"""
+
+from openedx_events.learning.data import ExamAttemptData, UserData, UserPersonalData
+from openedx_events.learning.signals import EXAM_ATTEMPT_SUBMITTED
+
+
+def emit_exam_attempt_submitted_event(user, course_key, usage_key, exam_type):
+    """
+    Emit the EXAM_ATTEMPT_SUBMITTED Open edX event.
+    """
+    user_data = UserData(
+        id=user.id,
+        is_active=user.is_active,
+        pii=UserPersonalData(
+            username=user.username,
+            email=user.email,
+            name=user.full_name
+        )
+    )
+
+    # .. event_implemented_name: EXAM_ATTEMPT_SUBMITTED
+    EXAM_ATTEMPT_SUBMITTED.send_event(
+        exam_attempt=ExamAttemptData(
+            student_user=user_data,
+            course_key=course_key,
+            usage_key=usage_key,
+            exam_type=exam_type,
+            requesting_user=user_data
+        )
+    )
--- a/edx_exams/apps/core/tests/test_api.py
+++ b/edx_exams/apps/core/tests/test_api.py
@@ -11,6 +11,7 @@ from django.conf import settings
 from django.utils import timezone
 from freezegun import freeze_time
 from opaque_keys.edx.keys import CourseKey, UsageKey
+from openedx_events.learning.data import ExamAttemptData, UserData, UserPersonalData
 
 from edx_exams.apps.api.test_utils import ExamsAPITestCase
 from edx_exams.apps.core.api import (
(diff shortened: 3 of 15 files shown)


edx_exams/apps/core/api.py
```diff
@@ -160,6 +170,24 @@ def _calculate_allowed_mins(exam):
...
+    STUDENT_REGISTRATION_COMPLETED.send_event(
+        user=UserData(
+            id=1,
+            is_active=True,
+            pii=UserPersonalData(
+                username=username,
+                email='test',
+                name='test',
+            )
+            # data=InstructorCompletionData(
```
MichaelRoytman, line 184: This was just me commenting what our event data will likely look like. I know that is still in flux and will change, though.

edx_exams/apps/core/tests/test_api.py
```diff
@@ -212,7 +213,7 @@ def setUp(self):
             resource_id=str(uuid.uuid4()),
             course_id=self.course_id,
             provider=self.test_provider,
-            content_id='abcd1234',
+            content_id='block-v1:edX+test+2023+type@sequential+block@1111111111',
```
MichaelRoytman, line 216: This is necessary because the `content_id` must be able to be parsed into a valid `UsageKey`.

alangsto (approved):
LGTM 👍

edx_exams/apps/core/tests/test_api.py
```diff
@@ -298,6 +299,33 @@ def test_submit_attempt(self):
             self.assertEqual(updated_attempt.status, ExamAttemptStatus.submitted)
             self.assertEqual(updated_attempt.end_time, timezone.now())
 
+    @patch('edx_exams.apps.core.signals.signals.EXAM_ATTEMPT_SUBMITTED.send_event')
```
ilee2u, line 302: Should this be in a separate file for testing signals specifically?

MichaelRoytman: The signal is emitted from the API, so I think that this test make the most sense in this file, because that's what I'm testing. I could move it into a separate class specifically for testing events emitted by the API functions, if you'd prefer?

If I were to have tests in `signals.py`, they would be to test that calling `emit_exam_attempt_submitted_event` results in an emitted signal. Do you think that's a useful test to have? I could add that.

ilee2u: Ah I see what you're saying, yeah it makes sense to do that test on the API level.

I guess I'm moreso trying to figure out if we need to do any testing on the level of the signal itself. I'm doing some digging now to see if that's necessary.

MichaelRoytman: Let me know what you find! I don't think we need to test the actual Django signal API. All the plumbing that funnels Django signals to the event bus is all behind the scenes and not something we could test, I believe. But we could test whether calling `emit_exam_attempt_submitted_event` results in an emitted signal.

ilee2u: Ok I found an example, there's a [test here](https://github.com/openedx/edx-platform/blob/e45460a2ce6777c6440a76c5fbb2181d5edfa436/docs/hooks/events.rst#L4) in edx-platform certificates that seems similar to yours.

That tests seems to go about as deep as yours does so I think it's good as is.

ilee2u (approved):

edx_exams/apps/core/signals/signals.py
```diff
@@ -0,0 +1,21 @@
...
+                id=user.id,
+                is_active=user.is_active,
+                pii=UserPersonalData(
+                    username=user.username,
+                    email=user.email,
+                    name=user.full_name
+                )
+            ),
+            course_key=course_key,
+            usage_key=usage_key,
```
alangsto, line 19: Is there also supposed to be a `requesting_user` passed along? Looking at https://github.com/openedx/openedx-events/blob/fec19895eed1c37e4da6305bfa9849665a832372/openedx_events/learning/data.py#L316

alangsto: Or I guess it defaults to None, and only a learner can submit their own exam

MichaelRoytman: I don't think so. The plan was to reserve this attribute for cases where a user other than the learner has made the change, like an instructor resetting an exam in the Instructor Dashboard or overriding an exam attempt status on manual review, for example. I don't believe that submission can be triggered by anyone but the learner.

MichaelRoytman: I went back on this and decided to add in the `requesting_user` whenever there is one. I would prefer to be explicit about the `requesting_user` than having an implicit understanding that it should be empty whenever the learner is the requester.

edx_exams/apps/core/signals/signals.py
```diff
@@ -0,0 +1,33 @@
...
+
+from openedx_events.learning.data import ExamAttemptData, UserData, UserPersonalData
+from openedx_events.learning.signals import EXAM_ATTEMPT_SUBMITTED
+
+
+def emit_exam_attempt_submitted_event(user, course_key, usage_key, exam_type):
+    """
+    Emit the EXAM_ATTEMPT_SUBMITTED Open edX event.
+    """
+    name = user.full_name or ''
```
MichaelRoytman, line 13: This is because the `full_name` field of the `User` model can either be the empty string (`blank=True`) or None (`null=True`). This comes from the edX [cookiecutter](https://github.com/openedx/edx-cookiecutters/blob/df10fdadc365f41e81bf2429c8b68ab13be4cb3c/cookiecutter-django-ida/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.project_name%7D%7D/apps/core/models.py#L17). I'm not sure what the rationale was for designing the model this way; setting both attributes to true is something [Django discourages](https://docs.djangoproject.com/en/4.2/ref/models/fields/#null). I created a thread in Slack to try to get some information on this decision. For now, I've done this. I'd prefer to fix the `User` model, but this will do until I have more information. I'll discuss this during standup.

Zacharis278 (approved):
👍

PR #180, opened 2023-09-22
Set up Kafka Open edX event bus.

MichaelRoytman:
**JIRA:** None.

**Description:** This pull requests adds necessary configuration and packages to run the Kafka implementation of the Open edX event bus.

**Author concerns:** None.

**Dependencies:** None.

**Testing instructions:**

None.

**Merge checklist:**
- [x] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/settings/devstack.py
+++ b/edx_exams/settings/devstack.py
@@ -52,3 +52,15 @@ JWT_AUTH.update({
 })
 
 LMS_ROOT_URL = 'http://edx.devstack.lms:18000'
+
+EVENT_BUS_PRODUCER = 'edx_event_bus_kafka.create_producer'
+EVENT_BUS_CONSUMER = 'edx_event_bus_kafka.KafkaEventConsumer'
+EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL = 'http://edx.devstack.schema-registry:8081'
+EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS = 'edx.devstack.kafka:29092'
+EVENT_BUS_TOPIC_PREFIX = 'dev'
+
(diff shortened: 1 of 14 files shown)


edx_exams/settings/devstack.py
```diff
@@ -52,3 +52,9 @@
 })
 
 LMS_ROOT_URL = 'http://edx.devstack.lms:18000'
+
+EVENT_BUS_PRODUCER = 'edx_event_bus_kafka.create_producer'
+EVENT_BUS_CONSUMER = 'edx_event_bus_kafka.KafkaEventConsumer'
+EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL = 'http://edx.devstack.schema-registry:8081'
+EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS = 'edx.devstack.kafka:29092'
+EVENT_BUS_TOPIC_PREFIX = 'dev'
```
ilee2u, line 60: Are we also supposed to add a blank value for `EVENT_BUS_PRODUCER_CONFIG` per the plan that Becca gave, or is that supposed to be done when we pushed to prod?

ilee2u (approved):

edx_exams/settings/devstack.py
MichaelRoytman: I think it makes sense to add here as well. Otherwise, it won't work in development. I'll add it.

We'll have to add all of this configuration to the stage and production configuration files at some point, which is where that setting will really matter.

ilee2u (approved):
LGTM pt. II

PR #187, opened 2023-09-29
docs: downstream effect events corrected

ilee2u:
Corrected this ADR to align with the hackathon doc: https://2u-internal.atlassian.net/wiki/spaces/PT/pages/577699957/M6+Hackathon+Outline#Event-Signals%2C-Their-Data%2C-and-Destinations%3A

### DIFF
--- a/docs/decisions/0004-downstream-effect-events.rst
+++ b/docs/decisions/0004-downstream-effect-events.rst
@@ -18,11 +18,10 @@ Decision
 We've decided to use an event driven architecture because unlike the legacy system (edx-proctoring), edx-exams is an independent service.
 As such, we plan to use the event bus to send info to edx-platform services without needing a response as one would in a REST framework.
 
-Below are lists of downstream effects of exam submission and review that we will or will not be impementing as downstream effects that will be
-triggered by events emitted in edx-exams per this decision.
+Below are lists of expected downstream effects of exam submission and review. We have decided to implement some downstream effects and not others; these decisions are enumerated below. Those that will be implemented will be triggered by events emitted by the edx-exams service, as described in this decision.
 
-Downstream effects to be implemented:
-***********************************************
+Downstream effects to be implemented as part of this decision:
+**************************************************************
 
   * Grades Override - A python API call to the grades app to generate a grade override when an exam attempt is rejected.
 
@@ -30,7 +29,7 @@ Downstream effects to be implemented:
 
   * Instructor Delete Attempt - A python API call to the instructor app to delete an exam attempt.
 
-  * Instructor Complete Attempt - A python API call to the instructor app to mark an exam subsection as completed.
+  * Mark Exam Subsection Completed - A python API call to a function in the instructor app that causes the exam subsection to be marked as completed.
 
   * Invalidate Certificate - A python API call to the certificates app to invalidate a leaner’s edx certificate for a course.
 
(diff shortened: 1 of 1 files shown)


docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -15,12 +15,13 @@ which edx-proctoring can call directly since it is a plugin that is installed di
 
 Decision
 --------
-We've decided to use an event driven architecture because unlike the legacy system (edx-porctoring), edx-exams is an independent service.
+We've decided to use an event driven architecture because unlike the legacy system (edx-proctoring), edx-exams is an independent service.
 As such, we plan to use the event bus to send info to edx-platform services without needing a response as one would in a REST framework.
 
-Below are lists of downstream effects of exam submission and review that we will or will not be translating into events in edx-exams as part of this decision.
+Below are lists of downstream effects of exam submission and review that we will or will not be impementing as downstream effects that will be
```
MichaelRoytman, line 21:
```suggestion
Below are lists of downstream effects of exam submission and review that we will or will not be implementing as downstream effects that will be
```

MichaelRoytman: I'm having a hard time understanding this sentence. I've read it three times and am still struggling. Maybe it's because it's Friday afternoon.

Are you trying to say something like this?

"Below are lists of expected downstream effects of exam submission and review. We have decided to implement some downstream effects and not others; these decisions are enumerated below. Those that will be implemented will be triggered by events emitted by the `edx-exams` service, as described in this decision."

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -56,21 +57,40 @@ Downstream effects that are not being implemented as part of this decision:
 
     * We will implement this later since we do not believe this to be an essential feature.
 
-For easier visualization, here are all of the events we plan to implement, described from end to end:
+For easier visualization, here are all of the downstream effects we plan to port, described from end to end:
+
+Downstream effects to be implemented as part of this decision:
```
Zacharis278, line 62: NIT: after seeing this table rendered. The general context / expected result columns are way more useful than the exact files/functions we are calling but get cutoff when viewing the page. I'd move those left so people can read them easily. TBH the file/function might not even be worth it if it doesn't really fit on the page.

Zacharis278 (approved):
layout nit but otherwise looks good

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -56,21 +57,40 @@ Downstream effects that are not being implemented as part of this decision:
...
+ ======================== ======================================== ======================================================================== 
+  Event Type               Context in which event is produced       Downstream effects triggered                                            
+ ======================== ======================================== ======================================================================== 
+  Exam Attempt Submitted   When an exam attempt is submitted.       Instructor Mark Subsection Completed, Set Credit Requirement Status     
+  Exam Attempt Rejected    When an exam attempt is rejected.        Set Credit Requirement Status, Grades Override, Invalidate Certificate  
+  Exam Attempt Verified    When an exam attempt is verified.        Set Credit Requirement Status, Undo Grades Override                     
+  Exam Attempt Errored     When exam attempt errors out.            Set Credit Requirement Status
```
MichaelRoytman, line 90:
```suggestion
  Exam Attempt Errored     When an exam attempt errors out.            Set Credit Requirement Status                                           
```

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -15,12 +15,13 @@ which edx-proctoring can call directly since it is a plugin that is installed di
...
-We've decided to use an event driven architecture because unlike the legacy system (edx-porctoring), edx-exams is an independent service.
+We've decided to use an event driven architecture because unlike the legacy system (edx-proctoring), edx-exams is an independent service.
 As such, we plan to use the event bus to send info to edx-platform services without needing a response as one would in a REST framework.
 
-Below are lists of downstream effects of exam submission and review that we will or will not be translating into events in edx-exams as part of this decision.
+Below are lists of downstream effects of exam submission and review that we will or will not be impementing as downstream effects that will be
+triggered by events emitted in edx-exams per this decision.
 
-Downstream effects to be implemented as events:
+Downstream effects to be implemented:
```
MichaelRoytman, line 24: I'm noticing a slight mismatch between the names in this section and the names in the `Downstream effects to be implemented as part of this decision` section.

docs/decisions/0004-downstream-effect-events.rst
```diff
@@ -56,21 +57,40 @@ Downstream effects that are not being implemented as part of this decision:
...
+  Grades Override                        When an exam attempt is rejected.                                                                lms/djangoapps/grades/signals.py             override_subsection_grade in api.py             When we need to override a grade from any service.                        A grade override object is created or modified in the grades service within the LMS.  
+  Undo Grades Override                   When an exam attempt is verified after previously being rejected, OR when it is deleted/reset.   lms/djangoapps/grades/signals.py             undo_override_subsection_grade in services.py   When we need to undo a grade override from any service.                   A grade override object is deleted in the grades service within the LMS.              
+  Instructor Reset Subsection            When an exam attempt is deleted/reset.                                                           lms/djangoapps/instructor/signals.py         reset_student_attempts in enrollments.py        When we need to reset a student’s state in a subsection.                  A learner's state for a subsection is reset.
```
MichaelRoytman, line 69: This is nit, but every time I see this, I read it as relating to the instructor dashboard reset feature, not the instructor service. This makes me read "Instructor Mark Subsection Completed the same way, even though it has nothing to do with the instructor dashboard. Do you think we need to keep "instructor" in these names, since the other downstream effects don't contain the related service names? I think it makes it confusing.

MichaelRoytman (approved):

PR #188, opened 2023-10-02
Alter User model full_name field to be non-nullable.

MichaelRoytman:
**JIRA:** None.

**Description:** 

This commit makes the `User.full_name` field non-nullable to ensure that there is only one representation of the empty value - the empty string. Currently, empty values for this field can be either None or the empty string. Because we're transitioning from a nullable to non-nullable state, a default is required to handle existing rows with a null value.

This is best practice; Django discourages setting both `null=True` and `blank=True` on `CharField` model fields. Furthermore, this was required by our event bus work. If an empty value is represented by None, this causes issues with the event bus, because None is not JSON serializable. Instead of converting a None value to the empty string in the event producer, correcting the the model definition is a better approach.

**Author concerns:** This is the default model definition in the [edX cookiecutter](https://github.com/openedx/edx-cookiecutters/blob/df10fdadc365f41e81bf2429c8b68ab13be4cb3c/cookiecutter-django-ida/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.project_name%7D%7D/apps/core/models.py#L17). This pattern has been propagated to other applications. I have been unable to find a justification for the choice to set `null=True` and `blank=True=, but there is the small potential this choice was meaningful.

**Dependencies:** None.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [ ] Changelog record added
- [x] Commits are squashed

**Post merge:**


ilee2u (approved):
LGTM 🚀

PR #192, opened 2023-10-04
Modify local Kafka settings.

MichaelRoytman:
**JIRA:** None.

**Description:**

This commit set Django settings in the local settings file for event bus networking settings to to `None`. These settings are used by the local application server. This change is necessary because the event bus does not work outside of a Docker container, because the event bus is run through the devstack networking layer, which is inaccessible by the local application server. Because the local settings file is also used by Tutor, only the networking related settings are set to `None`. The remaining settings are left as-is, because they are sensible defaults. Tutor uses can override these settings as need be.

**Author concerns:** None.

**Dependencies:** None.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/settings/devstack.py
+++ b/edx_exams/settings/devstack.py
@@ -53,6 +53,8 @@ JWT_AUTH.update({
 
 LMS_ROOT_URL = 'http://edx.devstack.lms:18000'
 
+# EVENT BUS
+
 EVENT_BUS_PRODUCER = 'edx_event_bus_kafka.create_producer'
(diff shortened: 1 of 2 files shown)


Zacharis278 (commented):
wondering if these keys should still be here but the URLs are just 'None' and we can just comment on why.

Tutor is based on these local settings and will override the URLs but the rest of the settings are still useful

MichaelRoytman:
@Zacharis278 Are you thinking just the URLs are set to `None`? Or all the event bus settings?

Zacharis278 (approved):

PR #194, opened 2023-10-04
Document developing on the devstack Open edX event bus.

MichaelRoytman:
**JIRA:** None.

**Description:** This pull requests adds documentation for how to set up the Open edX event bus locally using devstack.

**Author concerns:** None.

**Dependencies:** None.

**Merge checklist:**
- [x] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [x] Delete working branch (if not needed anymore)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,9 @@
 version: "2.1"
 services:
   db:
-    image: mysql:5.6
+    # Oracle-packaged version includes a `linux/arm64/v8` version, needed for
+    # machines with Apple Silicon CPUs (Mac M1, M2)
+    image: mysql:8.0.33-oracle
     container_name: edx_exams.db
     environment:
       MYSQL_ROOT_PASSWORD: ""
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,7 +1,9 @@
 version: "2.1"
 services:
   db:
-    image: mysql:5.6
+    # Oracle-packaged version includes a `linux/arm64/v8` version, needed for
```
MichaelRoytman, line 4: This is necessary because the `JSONField` Django model field is only supported in MySQL version greater than or equal to `5.7`. This makes it impossible to run migrations because the models in `xblock-lti-consumer` use `JSONField`. Because we have migrated to version `8.0`, I've updated the base MySQL image to `8.0`. This particular image is copied from [devstack](https://github.com/openedx/devstack/blob/fa758e4e51aee3df8d69af975e26154c4cd70bab/docker-compose.yml#L229).

Zacharis278 (approved):

PR #198, opened 2023-10-10
fix: end assessment launch

Zacharis278:
**JIRA:** [MST-2159](https://2u-internal.atlassian.net/browse/MST-2159?atlOrigin=eyJpIjoiZjdkY2YzMThiMzhkNGNiNWFlMDhkYzY1M2ZhNzQ2NGYiLCJwIjoiaiJ9)

**Description:** The user_id passed to this function is used to lookup the initial launch data which is based on the user's database id instead of their external/anonymous id. 

**Author concerns:** Pretty sure this is the issue but hitting some problems with the IMS tools locally so I can't validate. Going to try this out on stage where things are working.

### DIFF
--- a/edx_exams/apps/lti/views.py
+++ b/edx_exams/apps/lti/views.py
@@ -292,7 +292,7 @@ def end_assessment(request, attempt_id):
 
     exam = attempt.exam
     resource_link_id = exam.resource_id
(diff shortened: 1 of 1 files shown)


MichaelRoytman (approved):
This makes sense as being the root cause.

PR #205, opened 2023-10-26
docs: proctored exam testing instructions

ilee2u:
**JIRA:** Link to JIRA ticket

**Description:** Describe in a couple of sentences what this PR adds

migrations, etc.

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,6 +59,36 @@ Devstack Set Up
 
 You can use the make targets defined in the ``Makefile`` to interact with the running ``edx-exams`` Docker containers.
 
+LTI Configuration for Local Development
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+In order to use edx-exams as the exams backend for your courses (and to get it to wor in general), you will also need to complete the following configuration steps:
+
+#. Add EXAMS_SERVICE_URL = 'http://host.docker.internal:18740/api/v1' to your cms and lms private.py environment.
+#. In your terminal, navigate to your 
+#. In your terminal, run the frontend-app-course-authoring MFE by navigating to the devstack folder and running large-u .
+#. Ensure that the exams IDA is enabled for your course. In your LMS' django admin's course waffle flag table (http://localhost:18000/admin/waffle_utils/waffleflagcourseoverridemodel/) add an entry:
+    #. Enter course_apps.exams_ida in the Waffle Flag field
+    #. Enter the course ID for the course ID field (e.g. something like course-v1:edX+COURSENAME+T2023
+    #. Override Choice should be Force On
+    #. Select the Enabled checkbox.
+#. At http://localhost:18740/admin/core/proctoringprovider/add/, add an entry for a proctoring provider:
+    #. Make the "name" something with no spaces, like "custom-proctor"
+    #. Set the "verbose name" to whatever you want, maybe something like "Custom Proctoring Service"
+    #. The LTI configuration id and Tech support phone & email are not relevant for local testing, so just set them all to "1”.
+#. From whatever root folder you keep your devstack folders in, navigate to edx-platform/cms/envs/
+    #. If you have not already, create a private.py file in this /envs/ directory
+    #. In this private.py file, add the line FEATURES['ENABLE_EXAM_SETTINGS_HTML_VIEW'] = True
+#. Again from whatever root folder you keep your devstack folders in, navigate to frontend-app-course-authoring/.env.development
+    #. Set the environment variable EXAMS_BASE_URL="http://localhost:18740"
+    #. You will need to reset the frontend-app-course-authoring MFE if it is running for this setting to work
+#. Setup the proctoring provider in your course in studio (settings dropdown > proctored exam settings)
+    #. Set proctored exams to enabled
+    #. Set the proctoring provider to the name you entered in step #3
+#. Create a new proctored exam: 13.3. Creating Proctored Exams — Building and Running an Open edX Course documentation 
+    #. The exam registration task that registers exams with the exams service requires that the subsection due date (i.e. exam due date) is set and/or the course end date is set. Be sure to do do this; otherwise, your exam will not be registered with the exams service.
+#. Publish the course
+#. Ensure your exam in successfully created in edx-exams at http://localhost:18740/admin/core/exam/
+
 Development Workflow
 ~~~~~~~~~~~~~~~~~~~~
 .. code-block::
--- a/README.rst
+++ b/README.rst
@@ -59,36 +59,6 @@ Devstack Set Up
 
 You can use the make targets defined in the ``Makefile`` to interact with the running ``edx-exams`` Docker containers.
 
-LTI Configuration for Local Development
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-In order to use edx-exams as the exams backend for your courses (and to get it to work in general), you will also need to complete the following configuration steps:
-
-#. Add ``EXAMS_SERVICE_URL = 'http://host.docker.internal:18740/api/v1'`` to your cms and lms private.py environment settings file.
-#. Add ``EXAMS_BASE_URL='http://localhost:18740'`` to .env.development in frontend-app-learning.
-#. In your terminal, run the frontend-app-course-authoring MFE by navigating to the devstack folder and running ``make dev.up.large-and-slow``.
-#. Ensure that the exams IDA is enabled for your course. In your LMS' django admin's course waffle flag table (http://localhost:18000/admin/waffle_utils/waffleflagcourseoverridemodel/) add an entry:
-    #. Enter ``course_apps.exams_ida`` in the Waffle Flag field
-    #. Enter the course ID for the course ID field (e.g. something like ``course-v1:edX+COURSENAME+T2023``)
-    #. Override Choice should be Force On
-    #. Select the Enabled checkbox.
-#. At http://localhost:18740/admin/core/proctoringprovider/add/, add an entry for a proctoring provider:
-    #. Make the "name" something with no spaces, like "custom-proctor"
-    #. Set the "verbose name" to whatever you want, maybe something like "Custom Proctoring Service"
-    #. The LTI configuration id and Tech support phone & email are not relevant for local testing, so just set them all to "1”.
-#. From whatever root folder you keep your devstack folders in, navigate to edx-platform/cms/envs/
-    #. If you have not already, create a private.py file in this /envs/ directory
-    #. In this private.py file, add the line ``FEATURES['ENABLE_EXAM_SETTINGS_HTML_VIEW'] = True``
-#. Again from whatever root folder you keep your devstack folders in, navigate to frontend-app-course-authoring/.env.development
-    #. Set the environment variable ``EXAMS_BASE_URL="http://localhost:18740"``
-    #. You will need to reset the frontend-app-course-authoring MFE if it is running for this setting to work
-#. Setup the proctoring provider in your course in studio (settings dropdown > proctored exam settings)
-    #. Set proctored exams to enabled
-    #. Set the proctoring provider to the name you entered in step #3
-#. Create a new proctored exam: 13.3. Creating Proctored Exams — Building and Running an Open edX Course documentation 
-    #. The exam registration task that registers exams with the exams service requires that the subsection due date (i.e. exam due date) is set and/or the course end date is set. Be sure to do do this; otherwise, your exam will not be registered with the exams service.
-#. Publish the course
-#. Ensure your exam in successfully created in edx-exams at http://localhost:18740/admin/core/exam/
-
 Development Workflow
 ~~~~~~~~~~~~~~~~~~~~
 .. code-block::
--- a/README.rst
+++ b/README.rst
@@ -61,14 +61,14 @@ You can use the make targets defined in the ``Makefile`` to interact with the ru
 
 LTI Configuration for Local Development
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-In order to use edx-exams as the exams backend for your courses (and to get it to wor in general), you will also need to complete the following configuration steps:
+In order to use edx-exams as the exams backend for your courses (and to get it to work in general), you will also need to complete the following configuration steps:
 
-#. Add EXAMS_SERVICE_URL = 'http://host.docker.internal:18740/api/v1' to your cms and lms private.py environment.
-#. In your terminal, navigate to your 
-#. In your terminal, run the frontend-app-course-authoring MFE by navigating to the devstack folder and running large-u .
+#. Add ``EXAMS_SERVICE_URL = 'http://host.docker.internal:18740/api/v1'`` to your cms and lms private.py environment settings file.
+#. Add ``EXAMS_BASE_URL='http://localhost:18740'`` to .env.development in frontend-app-learning.
+#. In your terminal, run the frontend-app-course-authoring MFE by navigating to the devstack folder and running ``make dev.up.large-and-slow``.
 #. Ensure that the exams IDA is enabled for your course. In your LMS' django admin's course waffle flag table (http://localhost:18000/admin/waffle_utils/waffleflagcourseoverridemodel/) add an entry:
-    #. Enter course_apps.exams_ida in the Waffle Flag field
-    #. Enter the course ID for the course ID field (e.g. something like course-v1:edX+COURSENAME+T2023
+    #. Enter ``course_apps.exams_ida`` in the Waffle Flag field
+    #. Enter the course ID for the course ID field (e.g. something like ``course-v1:edX+COURSENAME+T2023``)
     #. Override Choice should be Force On
     #. Select the Enabled checkbox.
 #. At http://localhost:18740/admin/core/proctoringprovider/add/, add an entry for a proctoring provider:
@@ -77,9 +77,9 @@ In order to use edx-exams as the exams backend for your courses (and to get it t
     #. The LTI configuration id and Tech support phone & email are not relevant for local testing, so just set them all to "1”.
 #. From whatever root folder you keep your devstack folders in, navigate to edx-platform/cms/envs/
     #. If you have not already, create a private.py file in this /envs/ directory
-    #. In this private.py file, add the line FEATURES['ENABLE_EXAM_SETTINGS_HTML_VIEW'] = True
+    #. In this private.py file, add the line ``FEATURES['ENABLE_EXAM_SETTINGS_HTML_VIEW'] = True``
 #. Again from whatever root folder you keep your devstack folders in, navigate to frontend-app-course-authoring/.env.development
-    #. Set the environment variable EXAMS_BASE_URL="http://localhost:18740"
+    #. Set the environment variable ``EXAMS_BASE_URL="http://localhost:18740"``
     #. You will need to reset the frontend-app-course-authoring MFE if it is running for this setting to work
 #. Setup the proctoring provider in your course in studio (settings dropdown > proctored exam settings)
     #. Set proctored exams to enabled
--- a/docs/test_plan/proctored_exam.rst
+++ b/docs/test_plan/proctored_exam.rst
@@ -62,7 +62,8 @@ Exam Flow Works With Proctoring Software Installed & Running
 #. Clicking "Yes end my proctored exam" should end the exam and directs you to a "You have submitted this proctored exam for review" page
 #. Clicking "Yes end my proctored exam" should also close your proctoring software automatically.
 #. In the course view, the exam you took should be marked as completed (with a green checkmark)
-
+#. We recommend that you repeat these steps several times using different users (which you may have to provision if developing locally)
+    - Alternatively, you can also create mock exam attempt data in your edx exams admin (If developing locally, this is located at http://localhost:18740/admin/core/examattempt/).
 
 Proctorio-Specific Tests
 ========================
@@ -76,7 +77,6 @@ Exam Flow Works as Expected
 #. In a separate exam, go through the setup again and on the "Sending you back to your exam." page, click "Return to Exam". Clicking this link should send you to another interstitial. (NOTE: Clicking this link is time-bound so make sure to click it fast)
 #. Clicking the "Return to Exam" button should direct you to a page titled "Important" w/ another header saying "Proctored Exam Rules", and a "Start Exam" Button
 
-
 Exams Dashboard
 ===============
 Expected Behavior: Calls to the ACS Endpoint work, instructors can review/reset exam attempts, and expected downstream effects (e.g. w/ completion, credits, certificates, etc) of modifying an exam attempt's status occur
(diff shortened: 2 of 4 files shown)


docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,130 @@
...
+#. Clicking on the "Special Exams" tab loads the *new* version of the exams dashboard.
+#. Clicking on the "Review Dashboard" tab within the exams dashboard should open a new tab to the Proctorio Dashboard for instructors.
+#. Clicking the "Select An Exam" dropdown shows the list of exams for the course.
+#. Filtering by exam name in the exam selection dropdown brings up the expected exams (e.g. querying "M4-M6" shows all the exams in that section)
+#. Selecting a specific exam (i.e. the ones you just took) loads the respective exam attempts for those exams
+#. The exam attempts are in the expected status, e.g. successful attempts are submitted or verified, and the errored attempts are errored.
+#. Filtering by username only shows exam attempts for that username.
+#. The status changes for these exam attempts look the same in the edx-exams admin
+
+Resetting an exam attempt
```
Zacharis278, line 69: these tests could be a good place to cover other effects of a reset too. Why not test the learner's completion and problem is reset as well here? (same comment goes for the below tests)

Zacharis278 (commented):
few comments to get started.

High level, it would be nice to make this less Proctorio specific. If we do have a few Proctorio specific cases we want to cover we keep that as a separate test. The user flow should _mostly_ be the same regardless of the tool used.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,126 @@
...
+        #. The exams IDA waffle flag can be found/created in the LMS' Django Admin under "Waffle Flag Course Overrides"
+        #. If needed, you can make a new entry using "course_apps.exams_ida" as the waffle flag, the course ID, and by setting it to "Enabled"
+    #. Find or create two exam subsections with a type of 'Proctored exam' in your test course
+        #. This can be done by creating a subsection and clicking the 'Configure' button (should look like a gear), going to the "Advanced Tab", and selecting "Proctored"
+        #. You MUST also set this subsection to be Graded as a Midterm or Final exam
+        #. This subsection should contain at least one unit with one gradeable block, e.g. a generic multiple choice question in order to test downstream effects to the gradebook
+    #. You should also have a tab open with your event bus logs
+
+
+Proctored Exam Flow
```
Zacharis278, line 26: I think these another set of cases here for submitting an attempt that is auto verified and one needs a second review.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,126 @@
...
+#. Clicking "Cancel" closes modal
+#. Clicking "Verify" marks attempt as verified both in the dashboard and in the edx-exams admin
+#. Clicking "Reject" marks attempt as reject both in the dashboard and in the edx-exams admin
+
+Resetting an exam attempt
+=========================
+#. Clicking "Reset" on an exam attempt brings up a confirmation modal
+#. Clicking "No (Cancel)" closes the modal and nothing changes
+#. Clicking "Yes I'm Sure" removes the exam attempt from the table (effectively confirming that it's been reset)
+#. The exam attempt no longer appears in the edx-exams admin
```
Zacharis278, line 125: NIT: this one part of the test requires additional access. Going back the the section as a learner and ensuring you get the correct interstitial can test the reset happened without needing admin. (the tester already has to login as a learner to check completion in the next step.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,126 @@
...
+-------------------
+Expected Behavior: "A paid track learner is able to start, complete, and submit a proctored exam using Proctoring Software that is launched via LTI."
+
+Without Proctoring Software Installed
+=====================================
+#. Enter a Proctored Exam without proctoring software installed. You should see an interstitial stating that "This exam is proctored" with a button that says "Continue to Proctored Exam"
+#. Clicking "Continue to Proctored Exam" should bring you to an interstitial titled "Set up and start your proctored exam." with "Start System Check" and "Start Exam" buttons
+#. Clicking the "Start Exam" button should pop up a modal that says "Cannot Start Exam"
+#. Clicking "OK" closes that modal
+#. Clicking "Start System Check" redirects you to a (likely broken) page
```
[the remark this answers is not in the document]
ilee2u: No this is me getting mixed up with the exams dashboard for the proctoring software being a broken link at the moment. This shouldn't be broken ideally.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Exam Flow Works With Proctoring Software Installed & Running
+============================================================
+#. Go through the setup again, this time without closing your proctoring software. Clicking "Start Exam" should start the exam
+#. Complete whatever question you put inside your exam w/ the correct answer. Clicking "End My Exam" should direct you to a "Are you sure you want to end your proctored exam?" interstitial
+#. Clicking "No I'd like to continue working" should send you back to the exam
+#. Clicking "Yes end my proctored exam" should end the exam and directs you to a "You have submitted this proctored exam for review" page
+#. In the course view, the exam you took should be marked as completed (with a green checkmark)
+
+
```
alangsto, line 55: Are there any tests for what to expect when an error occurs? For the proctoring flow, these all look like "happy path" tests.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,134 @@
...
+        #. If needed, you can make a new entry using "course_apps.exams_ida" as the waffle flag, the course ID, and by setting it to "Enabled"
+    #. Find or create two exam subsections with a type of 'Proctored exam' in your test course
+        #. This can be done by creating a subsection and clicking the 'Configure' button (should look like a gear), going to the "Advanced Tab", and selecting "Proctored"
+        #. You MUST also set this subsection to be Graded as a Midterm or Final exam
+        #. This subsection should contain at least one unit with one gradeable block, e.g. a generic multiple choice question in order to test downstream effects to the gradebook
+    #. You should also have a tab open with your event bus logs
+
+
+Proctored Exam Flow
+-------------------
```
alangsto, line 27: NIT: using a single dash makes the text appear smaller than the following test cases. If this is meant to be the overarching category, I feel like this text should be bigger.

docs/test_plan/proctored_exam.rst
ilee2u: Yes those should be here: https://github.com/edx/edx-exams/pull/205/files/d1455d82318fcaebb675be2eff36909ceee74c6e#diff-b5c1a36768f79123cc3158c9c56b1b72e50008252b380bec3d570b0a2ca0cc37R119. I'm realizing I need to revise these to say "As a student, error the exam, and as an admin, reset the exam attempt, then try the exam again as a student" so that we reflect what would happen IRL.

docs/test_plan/proctored_exam.rst
ilee2u: OOPS

docs/test_plan/proctored_exam.rst
Zacharis278: I still think we're missing these test cases ensuring a review by the proctoring provider is received and the attempt is updated. We kind of do that as a side effect of setup for testing the instructor dashboard but I think it's best to be explicit. If someone is just testing exam flow they should be validating this case.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,130 @@
...
+#. Immediately navigating back and trying to click "Start Exam" still doesn't allow you to start the exam.
+#. Navigating back to the Proctorio Instructions and refreshing on that page causes no issues
+#. Go through the Proctorio Setup for your Webcam, Desktop, etc. Upon setup completion, your webcam should appear to the right of the screen along with some controls. You should land on a page that says "Sending you back to your exam." with a link to "Return to Exam". Clicking this link should send you to another interstitial. (NOTE: Clicking this link is time-bound so make sure to click it fast)
+#. Clicking "Return to Exam" directs you to a page titled "Important" w/ another header saying "Proctored Exam Rules", and a "Start Exam" Button
+#. Clicking "Start Exam" directs you to the exam.
+#. Clicking "End My Exam" directs you to a "Are you sure you want to end your proctored exam?" Interstitial
+#. Clicking "No I'd like to continue working" sends you back to the exam.
+#. Clicking "Yes end my proctored exam" ends the exam and directs you to a "You have submitted this proctored exam for review" page
```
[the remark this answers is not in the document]
Zacharis278: This is mostly just a NIT but bumping this comment because I think it's worth calling out specifically.

docs/test_plan/proctored_exam.rst
Zacharis278: I see we're testing the exams dashboard for error attempts but to Alie's point shouldn't we have a test of an error from the "testing the exam flow" perspective? Something to test that the learner gets an error interstitial / the learner cannot access the exam content after an error.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
...
+- `Creating a Proctored Exam <https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/proctored_exams/pt_create.html#creating-a-proctored-exam>`_
+
+Prerequisite steps:
+    #. Use Google Chrome
+    #. Have a staff account
+    #. Have a non-staff learner account enrolled in the verified track for this course
+    #. As your non-staff user in your current window, and enroll in the course where you have set up your proctored exam
+    #. In an incognito window, login to your staff account and visit your Support Tools page and search in the email for your non-staff user. You should now see the option to change the enrollment track of that user. Change the enrollment track of the user to ""verified"" (choose any reason when prompted, it doesn't matter)
+    #. In that same incognito, still signed in as staff, navigate to the Instructor dashboard
+    #. Find or create a test course within Stage with proctoring enabled and an LTI proctoring provider chosen, certificates are enabled, and where the exams IDA waffle flag is enabled
```
MichaelRoytman, line 16: I think that the order of these prerequisite steps is a little confusing. For example, you're enrolling a non-staff learner account in "this course" before you have set up a course with proctoring. Based on your instructions, you may set up an entirely new course in which the learner is not enrolled.

I'd recommend breaking this up into "Proctoring Set Up Steps" and "User Set Up Steps" to make this clearer.

MichaelRoytman: Do you have a particular environment in mind here (e.g. stage)? Stage will have proctoring set up in a way that someone just starting in their devstack will not, so I think we should either not make references to a particular environment and link out to documentation for each environment (e.g. stage/production and development), or specify that this is just for stage/production.

MichaelRoytman: "certificates are enabled" is a little vague and could mean a variety of things. I think we should be more specific here. Do you mean "where there is a verified track" or something else?

MichaelRoytman: I know we don't have official documentation for this yet, but it would be good to clarify what "an LTI proctoring provider chosen" means - how does one do this or verify this? Depending on which environment this documentation applies to, you could just refer to the Proctoring Settings Page and/or reference our [Local Development and LTI Configuration](https://2u-internal.atlassian.net/wiki/spaces/PT/pages/256737327/Local+Development+and+LTI+Configuration) document for the development environment.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
...
+        #. The exams IDA waffle flag can be found/created in the LMS' Django Admin under "Waffle Flag Course Overrides"
+        #. If needed, you can make a new entry using "course_apps.exams_ida" as the waffle flag, the course ID, and by setting it to "Enabled"
+    #. Find or create two exam subsections with a type of 'Proctored exam' in your test course
+        #. This can be done by creating a subsection and clicking the 'Configure' button (should look like a gear), going to the "Advanced Tab", and selecting "Proctored"
+        #. You MUST also set this subsection to be Graded as a Midterm or Final exam
+        #. This subsection should contain at least one unit with one gradeable block, e.g. a generic multiple choice question in order to test downstream effects to the gradebook
+    #. You should also have a tab open with your event bus logs
```
MichaelRoytman, line 23: I would have no idea what this means if I wasn't already familiar with the event bus. How does one do this? For what it's worth, you don't make a reference to the logs again, so is this step really necessary?

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
+Proctored Exams
+===============
+
+Helpful ReadTheDocs Pages:
+
+- `Configuring a Proctoring Provider <https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/proctored_exams/proctored_enabling.html#configuring-proctoring-provider>`_
+- `Creating a Proctored Exam <https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/proctored_exams/pt_create.html#creating-a-proctored-exam>`_
```
MichaelRoytman, line 7: Neither of these links is used in this document.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
...
+----------------------------------------------------
+#. Install the proctoring software that you've configured with your platform. Go through the opening interstitials and click "Start System Check". This should load your respective Proctoring Software Setup page
+#. Immediately navigating back and trying to click "Start Exam" should't allow you to start the exam
+#. Navigating back to your proctoring software setup page and refreshing on that page should cause no issues
+#. Go through your proctoring software setup for your Webcam, Desktop, etc. Upon setup completion, you should be redirected back to an exam interstitial with just the "Start Exam" button
+#. Close your proctoring software and click "Start Exam". The exam should not start
+
+Exam Flow Works With Proctoring Software Installed & Running
+------------------------------------------------------------
+#. Go through the setup again, this time without closing your proctoring software. Clicking "Start Exam" should start the exam
```
MichaelRoytman, line 49: What setup? The steps in "Can't Start Exam Without Proctoring Software Running"?

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
...
+
+
+Exams Dashboard
+===============
+Expected Behavior: Calls to the ACS Endpoint work, instructors can review/reset exam attempts, and expected downstream effects (e.g. w/ completion, credits, certificates, etc) of modifying an exam attempt's status occur
+NOTE: These instructions currently do not account for downstream effects involving a learner's credit requirement status. We currently have no plan to include these
+
+Dashboard UI Loads Exam Attempts
+--------------------------------
+#. Clicking on the "Special Exams" tab should load the *new* version of the exams dashboard
```
MichaelRoytman, line 76: Where is this tab? How do I get to this tab? How do I know if I'm at the old or new exams dashboard?

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
...
+#. Clicking the "Select An Exam" dropdown should show the list of exams for the course
+#. Filtering by exam name in the exam selection dropdown should bring up the expected exams (e.g. querying "M4-M6" shows all the exams in that section)
+#. Selecting a specific exam (i.e. the ones you just took) should load the respective exam attempts for those exams
+#. The exam attempts should be in the expected status, e.g. successful attempts are submitted or verified, and the errored attempts are errored
+#. Filtering by username only should show exam attempts for that username
+#. Clicking the "Review Dashboard" link in the modal should load the review dashboard for your chosen Proctoring Software (e.g. Proctortrack or Proctorio)
+
+Can Reject/Verify "Review Required" attempts
+---------------------------------------------------
+INSTRUCTIONS: Go into another exam and be suspiscious. Block your webcam, leave for a few seconds, make strange noises, open a bunch of tabs, watch some youtube, open wikipedia, ask chatgpt for the answer to life, etc. Note you will need to do this a couple of times
```
MichaelRoytman, line 87: Out of curiosity, how does this work? Does Proctorio return a response this quickly? I know Proctortrack would take some time to return back a review.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
...
+
+#. Submit your suspiscious exam attempt, and check the exams dashboard. Assuming you were suspiscious enough, you should see a row that has a status of "Review Required" AND has a "Review Required" button for your attempt
```
MichaelRoytman, line 89: Would it be faster and less vendor specific to go into the admin dashboard and set the attempt's status to `second_review_required`? This also takes a way the need to be "suspicious enough" and avoids relying on the vendor's review process, which we don't need to be testing here. This applies to all the test cases that rely on a review coming in.

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
...
+    #. NOTE: This exam attempt may appear with the status "Satisfied" if you are using Proctorio. This means Proctorio's system is still analyzing your attempt. If this happens, just wait a bit and refresh the page until the status changes to "Verified"
+#. Clicking the "Review Required" button should spawn a modal that says "Update review status"
+#. Clicking "Cancel" should close the modal
+#. Clicking "Verify" should mark attempt as verified both in the dashboard
```
MichaelRoytman, line 93: Both in the dashboard and where else?

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
...
+Can Review Verified Attempts
+----------------------------
+INSTRUCTIONS: Go into another exam and don't be suspiscious. Keep your webcam on with your face in frame. Complete the exam
+
+#. An exam attempt with a status of "Verified" should appear and have a button/link at the end of their rows titled "Manual Review"
+    #. NOTE: This exam attempt may appear with the status "Satisfied" if you are using Proctorio. This means Proctorio's system is still analyzing your attempt. If this happens, just wait a bit and refresh the page until the status changes to "Verified"
+#. A modal should pop up upon clicking "Manual Review" that contains "Cancel" and "Reject" Buttons
+#. Clicking "Cancel" should close the modal
+#. Clicking "Reject" should change the attempt status to "rejected" in the UI
+#. Upon refreshing a filtering for this exam attempt, it is still marked as "rejected"
```
MichaelRoytman, line 107: What is this testing? Why don't we do this same step for "Can Reject/Verify "Review Required" attempts"?

docs/test_plan/proctored_exam.rst
```diff
@@ -0,0 +1,133 @@
...
+#. Check the Gradebook (Instructor Dashboard -> Student Admin -> Gradebook) and check that the grade for the exam is 0
+#. Check the Generated Certificates table in the LMS admin at "{your platform url}/admin/certificates/generatedcertificate/" to see if the user's certificate for the course has been marked as "invalidated"
+
+Can Review Rejected Attempts
+----------------------------
+#. The exam attempt that you just marked as "Rejected" should have a button at the end of their rows saying "Manual Review"
+#. A modal pops up upon clickin "Manual Review" that only should  ha the "Cancel" and "Verify" Buttons
```
MichaelRoytman, line 114:
```suggestion
#. A modal pops up upon clicking "Manual Review" that only should have the "Cancel" and "Verify" Buttons
```

MichaelRoytman (commented):
I know I've left a lot of feedback. I read these instructions from the perspective of a team member with limited familiarity with LTI for proctoring, Proctorio, and how to set up proctoring. I think there there is a good amount of assumed knowledge in this document that not everyone will have, and my comments reflect that. I'm not marking my feedback as blocking. I know it's a pain, but I do think this document could really benefit from being a lot more explicit.

docs/test_plan/proctored_exam.rst
ilee2u: - Changed the order of the steps as you recommended (still need to push the change)
- I have local in mind, but based on [this thread](https://twou.slack.com/archives/C049EF6VCTB/p1699386718501819) it seems like it's also fair to mention that we do have a test course in stage (at least as of now) 
- I'm honestly not too sure how certificates work so if "verified track" makes more sense I can just put that.
- Assuming `edx-exams` ends up in Open edX, would it make sense to just move the instructions for [Local Development and LTI Configuration](https://2u-internal.atlassian.net/wiki/spaces/PT/pages/256737327/Local+Development+and+LTI+Configuration) to the README and just make that the source of truth?

docs/test_plan/proctored_exam.rst
ilee2u: Ah yeah that's just something I forgot to delete. I agree with you on this so I'll remove it.

docs/test_plan/proctored_exam.rst
ilee2u: Ah these are copied over from [timed_exam.rst](https://github.com/edx/edx-exams/blob/main/docs/test_plan/timed_exam.rst), where it seems like they aren't used either. I see what you're saying though, I can have these links referenced in the setup instructions where it makes sense.

docs/test_plan/proctored_exam.rst
ilee2u: Usually takes no more than a minute from what I've seen on Stage.

docs/test_plan/proctored_exam.rst
ilee2u: Ah that's a typo

docs/test_plan/proctored_exam.rst
ilee2u: I'm realizing this test is redundant because the status change in the UI only occurs if the status change in the backend was successful (this is done via "dispatch" in the exam dashboard's thunks), so I'm just going to remove it.

docs/test_plan/proctored_exam.rst
ilee2u: Added instructions on how to get to this tab + how to tell if it's the newer one.

docs/test_plan/proctored_exam.rst
ilee2u: Yes :)

ilee2u:
Made some changes based off @MichaelRoytman and @Zacharis278's comments, still have some work to do though.

docs/test_plan/proctored_exam.rst
ilee2u: Added instructions to have this as an option to just retaking the exam

docs/test_plan/proctored_exam.rst
ilee2u: I added a test to the Proctorio section about checking that a review of the exam exists. Regardless of whether or not we're using Proctorio, wouldn't the test that we received a review by a proctoring provider be checking the exams dashboard itself?

docs/test_plan/proctored_exam.rst
Zacharis278: for the local instructions you are correct and those do need to move as part of the release to openedx. We've got that noted as a requirement here: https://2u-internal.atlassian.net/wiki/spaces/PT/pages/626753683/Beta+Release+and+Beyond.

I do think these tests need to be done in a deployed environment and not Tutor/Devstack in order to say we're confident in releasing these features.

docs/test_plan/proctored_exam.rst
Zacharis278: I think this might be how they are referenced the other docs has them right next to the step to create a proctored exam. Maybe we just want to point back to this if these instructions are needed anywhere later in the doc. If you use headings you can add internal links within the doc like this https://github.com/openedx/edx-proctoring/blob/master/docs/testing/test_plan.md?plain=1#L220. That can also be good to avoid repeating yourself  if one test depends on the setup of another

docs/test_plan/proctored_exam.rst
Zacharis278: @MichaelRoytman I do think there's also going to be a gap until we get instructions in readthedocs. I don't think we should duplicate how to setup a proctored exam, a user, a proctoring provider, etc here and should just reference readthedocs where needed.

docs/test_plan/proctored_exam.rst
Zacharis278: I think the test I'm looking for is something that validates the ACS works as intended. So a test along the lines of "if a user submits an attempt without violation the proctoring software can update the attempt to verified" / "if a user submits with violations the proctoring software can update the attempt to review required".

I think that's a distinctly different test from testing that a verified/rejected/second review attempt can be reset in the UI since that testing instructor experience but the ACS is part of the learner / exam flow.

docs/test_plan/proctored_exam.rst
MichaelRoytman: Ah, I see. So it's like reference materials? My bad - got confused!

docs/test_plan/proctored_exam.rst
MichaelRoytman: Yeah, totally agree with linking out. I didn't mean to suggest we should duplicate - just that I felt like some information was missing, which can definitely be added by referencing other documentation, like "Set up a proctored exam: <link>."

docs/test_plan/proctored_exam.rst
MichaelRoytman: Oh, nice!

MichaelRoytman (commented):
Thanks for addressing my first set of feedback.

I added some non-blocking feedback, which you can take or leave.

I agree with Zach regarding the ACS tests.

README.rst
```diff
@@ -59,6 +59,36 @@ Devstack Set Up
 
 You can use the make targets defined in the ``Makefile`` to interact with the running ``edx-exams`` Docker containers.
 
+LTI Configuration for Local Development
```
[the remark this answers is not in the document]
ilee2u: I'd say so. Since this is meant to be seen by Open Edx's community, I believe it makes sense for this public README to be the source of truth instead of our internal doc. I'll make sure to do that link once this is merged.

Zacharis278 (approved):
I'd say we're looking good. One quick thing I noticed, there's a parent readme for this folder that has a link to the other sets of tests. Can you add a link there? https://github.com/edx/edx-exams/tree/main/docs/test_plan

PR #208, opened 2023-11-03
feat: roster endpoint POC

Zacharis278:
**JIRA:** [COSMO-52](https://2u-internal.atlassian.net/browse/COSMO-52?atlOrigin=eyJpIjoiY2VhNTVjZGQzOGFmNGVkYzhiNTVhMTBhYTY3ZThmYmYiLCJwIjoiaiJ9)

**Description:** This is an experimental feature to test out an approach for passing a course roster to the Proctorio app. Long term it would be ideal that this app and any future apps use existing LTI endpoints such as the Names and Roles Provisioning endpoint to access this data. This one-off url is ultimately a shortcut to get us to releasable state faster.

Basically we make use of the same auth pattern we created for the start proctoring launch where we mint a new session token to be used in a third party UI flow since the correct headers to use default JWT auth won't be set. See #59


alangsto (approved):
👍

PR #223, opened 2023-12-07
feat: add custom domain param to LTI launch

Zacharis278:
**JIRA:** [COSMO-126](https://2u-internal.atlassian.net/browse/COSMO-126?atlOrigin=eyJpIjoiOTBiNTIyMGZkZDQ3NDVlNmJkYzY2M2MzNzU5NjFhZWEiLCJwIjoiaiJ9)

**Description:**

Adds a `custom_url` parameter to the LTI launch. We'll use this to test a few possible approaches to using this url. More details in [COSMO-105](https://2u-internal.atlassian.net/browse/COSMO-105)

I've added an entirely new setting for this since that gives us maximum flexibility to play around in stage but I expect the end solution will be a bit more elegant and just use the existing MFE url setting.


schenedx (approved):
👍

PR #226, opened 2023-12-14
fix: not all usage keys matching

Zacharis278:
Fixes a bug where the existing regex would not match the entire content id into the capture group

This would break exams for any course with a `.` in the course id (and likely several other cases).


edx_exams/apps/core/constants.py
```diff
@@ -15,6 +15,6 @@ class Status:
 
 COURSE_ID_PATTERN = rf'(?P<course_id>({INTERNAL_COURSE_KEY_PATTERN}|{EXTERNAL_COURSE_KEY_PATTERN}))'
 
-CONTENT_ID_PATTERN = r'(?P<content_id>([A-Za-z0-9-_:@\+]+))'
+USAGE_KEY_PATTERN = r'(?P<content_id>(?:i4x://?[^/]+/[^/]+/[^/]+/[^@]+(?:@[^/]+)?)|(?:[^/]+))'
```
Zacharis278, line 18: pulled from edx-platform. I updated the variable to keep the terminology for this matching string consistent.

MichaelRoytman (approved):

PR #227, opened 2023-12-15
feat: add proctorio redirect url

Zacharis278:
**JIRA:** [COSMO-105](https://2u-internal.atlassian.net/browse/COSMO-105)

**Description:**

`custom_url` represents the url seen in the browser during an exam. When the extension is installed navigating to this URL will activate it and iframe in the exam content. By using a url we own we can handle what happens when the extension is not installed since the GET to /browser_lock will hit this service. If that happens we redirect the learner to getproctorio.com to install the extension.

The downside of this approach is getproctorio.com can't send the user back into the exam since we don't own that page. The learner would need to manually navigate back the the section and click start system check again.

To make this flow a bit smoother we would prefer learners download the extension first, before ever clicking 'start system check'. This ticket [COSMO-130](https://2u-internal.atlassian.net/browse/COSMO-130?atlOrigin=eyJpIjoiOTllMTliMDUxZTkzNDZmZTliYjlkMTE3YzkzMjYxYWYiLCJwIjoiaiJ9) adds that instruction to the interstitial. If they miss or ignore that instruction this PR will redirect them on the download page either way.

Note: We enforce the domain of messages from the extension. That will need to update along with this https://github.com/openedx/frontend-lib-special-exams/pull/126

### DIFF
--- a/edx_exams/apps/core/admin.py
+++ b/edx_exams/apps/core/admin.py
@@ -3,8 +3,7 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
 from django.utils.translation import gettext_lazy as _
-
-from edx_exams.apps.lti.utils import get_lti_root
+from lti_consumer.utils import get_lti_api_base
 
 from .models import (
     AssessmentControlResult,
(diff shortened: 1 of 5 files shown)


edx_exams/apps/lti/views.py
```diff
@@ -235,12 +235,12 @@ def start_proctoring(request, attempt_id):
     lti_config = LtiConfiguration.objects.get(id=lti_config_id)
 
     proctoring_start_assessment_url = urljoin(
-        get_lti_root(),
+        get_lti_api_base(),
```
Zacharis278, line 238: this helper function already existed in the lti library. Removed a 'duplicate' util from this service

edx_exams/apps/core/admin.py
```diff
@@ -61,7 +61,10 @@ class ExamAttemptAdmin(admin.ModelAdmin):
 
     def change_view(self, request, object_id, form_url='', extra_context=None):
         extra_context = extra_context or {}
-        extra_context['LTI_ROOT'] = get_lti_root()
+        # technically this should be be using get_lti_view_base() but the
```
Zacharis278, line 64: the customization on this django view can eventually be removed. It was only needed when we didn't have a UI to test with. I didn't want to add all that to this PR tho

alangsto (approved):
👍

PR #228, opened 2023-12-18
feat: change proctorio redirect to static page

Zacharis278:
Follow up to #227 which didn't work.

#### Problem
The request I thought was getting blocked/replaced by the extension still goes through and because it's a 302 response it hijacks the whole page and redirects before the extension can launch. The result is you get stuck redirecting to getproctorio.com always and can't enter the exam.

#### Solution
We need to use a static page and either have the learner click a link or somehow detect the extension and redirect via javascript. I'm working with Proctorio on what the best path forward would be there but this should fix the redirect problem.

Note: I added some temporary body content to this page just to see how it is or isn't rendered. It _could_ cause some problems but there's no way to test this locally.


edx_exams/templates/proctorio_download.html
```diff
@@ -0,0 +1 @@
+loading Proctorio extension...
```
alangsto, line 1: Nit: newline

alangsto (approved):
One nit, otherwise LGTM

PR #250, opened 2024-02-22
fix: replace username with full name in attemp emails

ahtesham-quraish:
**JIRA:** 
[VAN-1833](https://2u-internal.atlassian.net/browse/VAN-1833)

**Description:** 
Remove username and change subject in attempts emails 

**Testing instructions:**
It has been tested locally 

| Before  | After |
| ------------- | ------------- |
|  image | image  |
|  image | image|

### DIFF
--- a/edx_exams/apps/core/email.py
+++ b/edx_exams/apps/core/email.py
@@ -26,10 +26,13 @@ def send_attempt_status_email(attempt, escalation_email=None):
 
     if attempt.status == ExamAttemptStatus.submitted:
         email_template = 'email/proctoring_attempt_submitted.html'
+        email_subject = 'Proctoring attempt submitted'
     elif attempt.status == ExamAttemptStatus.verified:
         email_template = 'email/proctoring_attempt_verified.html'
+        email_subject = 'Proctoring attempt verified'
     elif attempt.status == ExamAttemptStatus.rejected:
         email_template = 'email/proctoring_attempt_rejected.html'
+        email_subject = 'Proctoring attempt rejected'
     else:
         return  # do not send emails for other statuses
 
@@ -44,7 +47,6 @@ def send_attempt_status_email(attempt, escalation_email=None):
         contact_url = f'{settings.LMS_ROOT_URL}/support/contact_us'
         contact_url_text = contact_url
 
-    email_subject = f'Proctored exam {exam.exam_name} for user {attempt.user.username}'
     body = email_template.render({
         'exam_name': exam.exam_name,
         'course_url': course_url,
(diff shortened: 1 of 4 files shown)


rijuma (approved):
LGTM

edx_exams/apps/core/email.py
```diff
@@ -44,12 +44,13 @@ def send_attempt_status_email(attempt, escalation_email=None):
         contact_url = f'{settings.LMS_ROOT_URL}/support/contact_us'
         contact_url_text = contact_url
 
-    email_subject = f'Proctored exam {exam.exam_name} for user {attempt.user.username}'
+    email_subject = f'Proctored exam {exam.exam_name} for user {attempt.user.full_name}'
```
Zacharis278, line 47: AFAIK this IDA isn't populating full_name when it creates new User objects since up until this point there's been no use for it. I think you'll need to ensure that happens and possibly migrate that value into existing users.

Zacharis278 (dismissed):

edx_exams/apps/core/email.py
zainab-amir: Can you share how data is populated in this IDA?

Zacharis278: I'm not exactly sure, it's part of the cookie cutter installation for all IDAs. I think it's baked into JWT authentication within https://github.com/openedx/edx-drf-extensions somehow. Although, I imagine this could be an issue for all other IDAs as well depending on if the particular bit of code is looking at JWT details vs the User object from the database. This should be on the JWT but doesn't seem to always exist in the User model.

syedsajjadkazmii (approved):

MichaelRoytman (approved):
This looks good. Could you please update the commit message and pull request title to accurately reflect the changes, since we are not longer replacing the username with the full name?

ahtesham-quraish:
> This looks good. Could you please update the commit message and pull request title to accurately reflect the changes, since we are not longer replacing the username with the full name?

I made the change now

edx_exams/apps/core/email.py
```diff
@@ -44,7 +47,6 @@ def send_attempt_status_email(attempt, escalation_email=None):
         contact_url = f'{settings.LMS_ROOT_URL}/support/contact_us'
         contact_url_text = contact_url
 
-    email_subject = f'Proctored exam {exam.exam_name} for user {attempt.user.username}'
```
alangsto, line 47: Why can we not leave the exam name in?

ahtesham-quraish: Actually we have discussed subject line with UX team and they suggested us that we can remove the exam name from subject but if you still think we should get it back we can then discuss with UX again. Thanks

alangsto (approved):

PR #251, opened 2024-02-27
feat: handle instructor permission events

Zacharis278:
**JIRA:** https://2u-internal.atlassian.net/browse/COSMO-151 (2u-internal)

**Description:** Handles CourseAccessRole events that govern access to the exams dashboard and instructor API actions on exams.

Tested this against my local kafka event bus

### DIFF
--- a/edx_exams/apps/core/signals/handlers.py
+++ b/edx_exams/apps/core/signals/handlers.py
@@ -105,7 +105,7 @@ def listen_for_course_access_role_added(sender, signal, **kwargs):  # pylint: di
     if role not in COURSE_STAFF_ROLES:
         return
 
-    user, created = User.objects.get_or_create(  # pylint: disable=unused-variable
+    user, _ = User.objects.get_or_create(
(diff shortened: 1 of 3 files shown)


edx_exams/apps/core/signals/handlers.py
```diff
@@ -83,3 +90,47 @@ def listen_for_exam_attempt_reset(sender, signal, **kwargs):  # pylint: disable=
...
+    """
+    event_data = kwargs['course_access_role_data']
+    user_data = event_data.user
+    course_key = event_data.course_key
+    role = event_data.role
+
+    if role not in COURSE_STAFF_ROLES:
+        return
+
+    user, created = User.objects.get_or_create(  # pylint: disable=unused-variable
```
alangsto, line 108: Instead of adding the pylint disable command, could you just have `user, _ = ...`, or is there a reason we should have a variable for created?

alangsto (approved):
One nit, but otherwise looks good!

PR #253, opened 2024-02-29
fix: add local django constraint

UsamaSadiq:
## Description
- Added local Django constraint so the global Django constraint in edx-lint can be upgraded without any issue.
- Related Issue https://github.com/edx/edx-arch-experiments/issues/552


iamsobanjaved (approved):

iamsobanjaved:
It is not needed anymore, as the Django version has already upgraded to 4.2

PR #255, opened 2024-03-05
import course staff management command

varshamenon4:
**JIRA:** [COSMO-200](https://2u-internal.atlassian.net/browse/COSMO-200)

**Description:** Create a management command to populate the CourseStaffRole model and any related User objects with data from the LMS. Once merged create an ESRE ticket to upload the CSV file and run the command.


edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,72 @@
...
+            self.add_course_staff(row.get('username'), row.get('email'), row.get('role'))
+
+    def add_course_staff(self, username, email, role):
+        # todo: add error handling
+        user, _ = User.objects.get_or_create(
+            username=username,
+            email=email,
+        )
+        # todo: get course, where to get course?
+        course_id = ''
```
varshamenon4, line 59: [question] this should come from the LMS correct?

varshamenon4 (commented):
Just wanted to put this draft out for initial review to get feedback and to make sure I'm on the right track. Also, I'm guessing for testing that I can just create a CSV and test locally that this populates the db in django admin?

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
Zacharis278: yeah this would also have to come from the CSV. Which is just pulling this model https://github.com/openedx/edx-platform/blob/3a343f8301f71878628ad4e36e6160bdbec4862d/common/djangoapps/student/models/user.py#L1044

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,72 @@
...
+
+        logger.info('Bulk add course staff complete!')
+
+    def add_course_staff_from_csv(self, csv_file):
+        """
+        Add the given set of course staff provided in csv
+        """
+        reader = list(unicodecsv.DictReader(csv_file))
+        for row in reader:
+            self.add_course_staff(row.get('username'), row.get('email'), row.get('role'))
```
Zacharis278, line 50: you'll want to build up blocks of multiple rows and do a bulk add to avoid creating a ton of database transactions.

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,103 @@
...
+            dest='csv_path',
+            required=False,
+            help='Path to CSV file.')
+        parser.add_argument(
+            '--batch-size',
+            type=int,
+            default=10000,
+            dest='batch_size',
+            help='Batch size')
+        # todo: should there be a batch delay?
```
Zacharis278, line 42: 👍 I think this is a good idea

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def add_course_staff_from_csv(self, csv_file, batch_size, batch_delay):
+        """
+        Add the given set of course staff provided in csv
+        """
+        reader = list(unicodecsv.DictReader(csv_file))
+        for i in range(0, len(reader), batch_size):
+            CourseStaffRole.objects.bulk_create(
+                CourseStaffRole(
+                    user=User.objects.get_or_create(
```
Zacharis278, line 72: this looks like it's still going to create users 1 by 1 in a loop. I think you might need to break the user creation part out and do that first and roles second.

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
```diff
@@ -0,0 +1,96 @@
...
+    CourseStaffRole,
+    User,
+)
+from edx_exams.apps.core.test_utils.factories import CourseStaffRoleFactory, UserFactory
+
+# todo: is this the correct logger name?
+LOGGER_NAME = 'edx_exams.apps.core.management.commands.bulk_add_course_staff'
+
+
+class TestBulkAddCourseStaff(TestCase):
```
Zacharis278, line 20: I think it would be worth adding a test here that ensures this command produces the expected number or queries w/ `assertNumQueries`. example: https://github.com/edx/edx-exams/blob/a6bedee649c974ac6244472293bb159249ed7e77/edx_exams/apps/lti/tests/test_views.py#L716

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: Added!

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
varshamenon4: That makes sense! When I used this method, I see that there are a number of calls: https://github.com/edx/edx-exams/actions/runs/8394389300/job/22991407617?pr=255. Is it better to just check relative number of queries (for example, that batch size affects the number of queries for example) or is there a better way to finetune? I couldn't figure out a good way to constrain to only a certain type of query (for example, just the bulk query). Was poking around here: https://docs.djangoproject.com/en/5.0/topics/testing/tools/

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,96 @@
...
+                self.add_course_staff_from_csv(csv_file, batch_size, batch_delay)
+
+        logger.info('Bulk add course staff complete!')
+
+    def add_course_staff_from_csv(self, csv_file, batch_size, batch_delay):
+        """
+        Add the given set of course staff provided in csv
+        """
+        reader = list(unicodecsv.DictReader(csv_file))
+        # todo: filter by users that haven't been created
```
varshamenon4, line 66: Will remove this comment, but just wanted to call out how I am filtering out users that have already been created. Lemme know if there is anything I can do differently. It did seem that using exists() was the most efficient way to check this.

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
```diff
@@ -0,0 +1,77 @@
...
+    def test_add_course_staff_with_existing_user(self):
+        lines = ['amy,amy@pond.com,staff,course-v1:edx+test+f19\n']
+        with NamedTemporaryFile() as csv:
+            csv = self._write_test_csv(csv, lines)
+            with self.assertNumQueries(2):
+                call_command(self.command, f'--csv_path={csv.name}')
+                assert CourseStaffRole.objects.filter(user=self.user.id).exists()
+
+    def test_add_course_staff_with_new_user(self):
+        lines = ['pam,pam@pond.com,staff,course-v1:edx+test+f20\n']
```
varshamenon4, line 54: I can try to refactor this to avoid repetition if that is better. Wasn't sure if it was more readable to have repeating lines variable for each test.

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
```diff
@@ -0,0 +1,77 @@
...
+    def test_add_course_staff_with_batch_size(self):
+        lines = ['pam,pam@pond.com,staff,course-v1:edx+test+f20\n',
+                 'sam,sam@pond.com,staff,course-v1:edx+test+f20\n']
+        with NamedTemporaryFile() as csv:
+            csv = self._write_test_csv(csv, lines)
+            with self.assertNumQueries(2):
+                call_command(self.command, f'--csv_path={csv.name}', '--batch_size=1')
+                assert CourseStaffRole.objects.filter(course_id='course-v1:edx+test+f20').count() == 2
+
+    def test_add_course_staff_with_batch_delay(self):
```
varshamenon4, line 70: I'm just testing here that this works with the batch delay being different and there is no error, as I figured the testing of the delay itself wouldn't be in the scope of this. Open to feedback here though.

varshamenon4 (commented):
Ready for review! Added some specific questions about batch testing and also testing using assertNumQueries.

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
Zacharis278: You can probably just make this it's own test rather than baking it into the others as an extra assertion. For that one I'd try a bunch of extra rows to ensure the query count has not increased in an unexpected way based on more data.

For getting the expected number you should be able to walk though the code and count things up for what you expect. If it's different, it would be good to know why or where the extra query is coming from. There's a few ways to debug django queries but an easy one is to use

```
from django.db import connection
print(connection.queries)
```

Zacharis278 (commented):
still looking over things but a few comments to start

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: Oh cool! Yes, Michael did share this code as an example here: [manual_verifications](https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/verify_student/management/commands/manual_verifications.py#L100) management command. But I didn't understand that using __in will help it be more efficient! So I'll update to make more similar to this example.

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
```diff
@@ -0,0 +1,77 @@
...
+                call_command(self.command, f'--csv_path={csv.name}')
+                assert CourseStaffRole.objects.filter(user=self.user.id).exists()
+
+    def test_add_course_staff_with_new_user(self):
+        lines = ['pam,pam@pond.com,staff,course-v1:edx+test+f20\n']
+        with NamedTemporaryFile() as csv:
+            csv = self._write_test_csv(csv, lines)
+            with self.assertNumQueries(2):
+                call_command(self.command, f'--csv_path={csv.name}')
+                assert CourseStaffRole.objects.filter(course_id='course-v1:edx+test+f20').count() == 1
```
Zacharis278, line 59: it would be better to test if the correct user/role was created here rather than assert any role exists. Same for the below tests

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
Zacharis278: makes sense

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,98 @@
...
+    help = """
+    Add Course Staff in bulk from CSV.
+    Expects that the data will be provided in a csv file format with the first row
+    being the header and columns being: username, email, role, course_id.
+    Example:
+        $ ... bulk_add_course_staff --csv_path=foo.csv
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
```
MichaelRoytman, line 27: Should this argument be optional? I see that this management command is a no-op if you don't provide a CSV path,  but it suggests that something happened when it didn't (i.e. "Bulk add course staff complete!"). What do you think of either requiring the argument and making it a positional one or logging a different message when no CSV path is provided that indicates that it was a no-op?

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,98 @@
...
+            type=int,
+            default=200,
+            dest='batch_size',
+            help='Batch size')
+        parser.add_argument(
+            '--batch_delay',
+            type=float,
+            default=1.0,
+            dest='batch_delay',
+            help='Time delay in each iteration')
```
MichaelRoytman, line 44: It would be helpful to include the unit here.

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,98 @@
...
+            with open(csv_path, 'rb') as csv_file:
+                self.add_course_staff_from_csv(csv_file, batch_size, batch_delay)
+
+        logger.info('Bulk add course staff complete!')
+
+    def add_course_staff_from_csv(self, csv_file, batch_size, batch_delay):
+        """
+        Add the given set of course staff provided in csv
+        """
+        reader = list(unicodecsv.DictReader(csv_file))
```
MichaelRoytman, line 65: What made you choose the `unicodecsv` package? I ask because Python's native `csv` module has a `DictReader` class that should provide the same functionality. Looking at the docs for `unicodecsv`, it looks like it improves upon some unicode issues in Python 2, but we're on Python 3 here, so I'd imagine we don't benefit from `unicodecsv`. Where possible, I think it would be better to leverage native packages. I think this should be an easy swap without any impact to the command or its tests. What do you think?

As a nit, at least for the `csv.DictReader` class, the class is iterable, so you don't have to cast the value to a list; you can just iterate the instance. Maybe `unicodecsv.DictReader` is different? Oh, I see. You're casting it to a list because you're iterating `reader` twice and don't want to exhaust the iterable, right?

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,98 @@
...
+        users_to_create = []
+        users_existing = {u.username for u in User.objects.filter(username__in=[r.get('username') for r in reader])}
```
MichaelRoytman, line 67: Nice use of a set for a quick membership check!

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,98 @@
...
+        for row in reader:
+            # if not User.objects.filter(username=row.get('username')).exists():
```
MichaelRoytman, line 69: Do you need this commented out portion?

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+                for user in users_to_create[i:i + batch_size]
+            )
+            time.sleep(batch_delay)
+
+        # bulk create course staff
+        for i in range(0, len(reader), batch_size):
+            CourseStaffRole.objects.bulk_create(
+                CourseStaffRole(
+                    user=User.objects.get(
```
MichaelRoytman, line 89: I'm not sure about the answer to this question, and maybe this is not possible. But I noticed that you create the `users_existing` set querying the `User` model via the `username` field. Here, though, you're querying the `User` model via the `username` _and_ the `email` field. I'm wondering whether a scenario is possible where a user's would be found in `users_existing` but not in this query.

If we're pulling this data from the LMS, I think this could happen, because `email` is modifiable, but `username` isn't. I think `email` is maintained on our `User` model via syncing with the `JWT` when they hit `edx-exams` in the process of an exam, right? Do you think this is something we need to address here? I'm just not sure this code guarantees a user will exist, which could cause an exception here.

Since the function is in a transaction, it should be okay, because it'll rollback, we can correct the CSV, and try again, but it would be better to handle explicitly.

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
```diff
@@ -0,0 +1,114 @@
...
+            user=user[0].id,
+            course_id=course_id,
+            role=course_role,
+        ).exists()
+
+    def test_empty_csv(self):
+        lines = []
+        with NamedTemporaryFile() as csv:
+            csv = self._write_test_csv(csv, lines)
+            call_command(self.command, f'--csv_path={csv.name}')
```
MichaelRoytman, line 53: Should we assert something here, like no `CourseStaffRole` was created for `self.user`?

MichaelRoytman (commented):
This looks good! I left a few, hopefully quick, questions for you.

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: You're right, this should probably be required! I had referenced code from a couple of places and it must have been not required elsewhere. But for this case, it makes sense to require.

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: Ah, good point! I've changed to use csv. The only reason I used unicode was because I had referenced code from elsewhere... so not really a good reason haha. 

I primarily cast to a list because I check the length of the reader (in order to batch the creates for the course staff). Lemme know if this is a bad idea though!

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
varshamenon4: Good call, just added that!

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: Oooh, that's a really good point. Since the username is unique, could I just get by username instead? I've updated to be that way but let me know if that could cause any issues.

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: oops! just removed!

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
MichaelRoytman: I think that's good! `username` should not change.

MichaelRoytman (approved):
This looks great! Thank you for incorporating my feedback.

Zacharis278 (approved):
🚀

PR #256, opened 2024-03-11
feat: add consumer config

alangsto:
**JIRA:** [COSMO-205](https://2u-internal.atlassian.net/browse/COSMO-205)

**Description:** Now that openedx-events is above 8.6.0, events can be produced without using a secondary handler.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/settings/local.py
+++ b/edx_exams/settings/local.py
@@ -151,8 +151,17 @@ EVENT_BUS_TOPIC_PREFIX = 'dev'
 EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL = None
 EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS = None
 
-# The EVENT_BUS_PRODUCER_CONFIG Django setting introduced in openedx-events 8.6.0 does not work properly.
-# In order to unblock development, EVENT_BUS_PRODUCER_CONFIG should be set to the empty object. Signal handlers
-# should be added as before (i.e. versions <8.6.0).
-# Once openedx-events 8.6.0 is corrected, this Django setting should be set.
-EVENT_BUS_PRODUCER_CONFIG = {}
+attempt_submitted_event_config = EVENT_BUS_PRODUCER_CONFIG['org.openedx.learning.exam.attempt.submitted.v1']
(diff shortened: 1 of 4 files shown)


edx_exams/settings/local.py
```diff
@@ -151,8 +151,17 @@
 EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL = None
 EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS = None
 
-# The EVENT_BUS_PRODUCER_CONFIG Django setting introduced in openedx-events 8.6.0 does not work properly.
-# In order to unblock development, EVENT_BUS_PRODUCER_CONFIG should be set to the empty object. Signal handlers
-# should be added as before (i.e. versions <8.6.0).
-# Once openedx-events 8.6.0 is corrected, this Django setting should be set.
-EVENT_BUS_PRODUCER_CONFIG = {}
+attempt_submitted_event_config = EVENT_BUS_PRODUCER_CONFIG['org.openedx.learning.exam.attempt.submitted.v1']
```
varshamenon4, line 154: [question] could you clarify the difference between the config in local and base? Wondering why it is set in both places, but I might be missing something. Thanks!

varshamenon4 (commented):
lgtm! Just one clarifying question. Thanks!

edx_exams/settings/local.py
alangsto: in `base.py` the event producers are disabled, whereas for local development we want them enabled.

ilee2u (approved):
I'm guessing this will be easiest to test on stage, so lgtm!

alangsto:
@ilee2u I tested locally and these updates work! So testing on stage is next.

edx_exams/settings/local.py
varshamenon4: Oh I understand now with the other PR. This is set in edx-internal but we still need the settings set locally.

varshamenon4 (approved):
lgtm!

PR #288, opened 2024-07-15
feat: GET allowances

Zacharis278:
**JIRA:** [COSMO-368](https://2u-internal.atlassian.net/browse/COSMO-368)

**Description:** Adds API to retrieve a list of allowances by course id

The story suggested using generic DRF views for this since there's no business logic. I tried that out but found would introduce a new pattern without making it much simpler really. It would also mess with our class for exception logging

### DIFF
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -1800,7 +1800,7 @@ class AllowanceViewTests(ExamsAPITestCase):
         # course staff has access
         course_staff_user = UserFactory.create()
         CourseStaffRole.objects.create(user=course_staff_user, course_id=self.course_id)
-        response = self.get_api(course_staff_user, self.course_id)
+        response = self.request_api('get', course_staff_user, self.course_id)
         self.assertEqual(response.status_code, 200)
 
     def test_get_allowances(self):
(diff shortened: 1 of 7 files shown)


edx_exams/apps/core/admin.py
```diff
@@ -109,6 +109,7 @@ class CourseStaffRoleAdmin(admin.ModelAdmin):
 @admin.register(StudentAllowance)
 class StudentAllowanceAdmin(admin.ModelAdmin):
     """ Admin configuration for the Student Allowance model """
+    raw_id_fields = ('user', 'exam')
```
Zacharis278, line 112: updated since the default (a dropdown) is unusable for this on prod/stage

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -1752,3 +1753,101 @@ def test_exam_provider(self):
...
+
+    def setUp(self):
+        super().setUp()
+
+        self.course_id = 'course-v1:edx+test+f19'
+        self.exam = ExamFactory(
+            course_id=self.course_id,
+        )
+
+    def request_api(self, method, user, course_id):
```
Zacharis278, line 1771: overbuilt in anticipation of post+delete endpoints

edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -1752,3 +1753,101 @@ def test_exam_provider(self):
...
+        response = self.request_api('get', user, self.course_id)
+        self.assertEqual(response.status_code, 403)
+
+        # course staff has access
+        course_staff_user = UserFactory.create()
+        CourseStaffRole.objects.create(user=course_staff_user, course_id=self.course_id)
+        response = self.request_api('get', course_staff_user, self.course_id)
+        self.assertEqual(response.status_code, 200)
+
+    def test_get_allowances(self):
```
schenedx, line 1806: Should we have a DB query count check? I want to ensure this API only call DB once or twice, for even 1000 returned records.

schenedx (approved):
👍

edx_exams/apps/api/v1/tests/test_views.py
Zacharis278: IMO we only need to check query counts if is there's something more interesting happening in the business logic or model that might hide DB issues. In this case it's a pretty straightforward use of `.filter()`

PR #294, opened 2024-07-24
fix: handle dupe course staff mgmt command

varshamenon4:
**JIRA:** [COSMO-234](https://2u-internal.atlassian.net/browse/COSMO-234)

**Description:** This PR updates the course staff role management command to use ignore_conflicts to handle duplicate/already existing users and course staff roles. This is more efficient, and more readable. I did do some research, test locally and find other usages to confirm that ignore_conflicts is supported in our environment.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/core/management/commands/bulk_add_course_staff.py
+++ b/edx_exams/apps/core/management/commands/bulk_add_course_staff.py
@@ -63,32 +63,34 @@ class Command(BaseCommand):
         Add the given set of course staff provided in csv
         """
         reader = list(csv.DictReader(csv_file))
-        users_to_create = []
-        users_existing = {u.username for u in User.objects.filter(username__in=[r.get('username') for r in reader])}
-        for row in reader:
-            if row.get('username') not in users_existing:
-                users_to_create.append(row)
-                users_existing.add(row.get('username'))
 
         # bulk create users
-        for i in range(0, len(users_to_create), batch_size):
+        for i in range(0, len(reader), batch_size):
             User.objects.bulk_create(
-                User(
-                    username=user.get('username'),
-                    email=user.get('email'),
-                )
-                for user in users_to_create[i:i + batch_size]
+                (User(
+                    username=row.get('username'),
+                    email=row.get('email'),
+                ) for row in reader[i:i + batch_size]),
+                ignore_conflicts=True,
+            )
+            CourseStaffRole.objects.bulk_create(
+                (CourseStaffRole(
+                    user=User.objects.get(username=row.get('username')),
+                    course_id=row.get('course_id'),
+                    role=row.get('role'),
+                ) for row in reader[i:i + batch_size]),
+                ignore_conflicts=True,
             )
             time.sleep(batch_delay)
 
         # bulk create course staff
-        for i in range(0, len(reader), batch_size):
-            CourseStaffRole.objects.bulk_create(
-                CourseStaffRole(
-                    user=User.objects.get(username=row.get('username')),
-                    course_id=row.get('course_id'),
-                    role=row.get('role'),
-                )
-                for row in reader[i:i + batch_size]
-            )
-            time.sleep(batch_delay)
+        # for i in range(0, len(reader), batch_size):
+        #     CourseStaffRole.objects.bulk_create(
+        #         (CourseStaffRole(
+        #             user=User.objects.get(username=row.get('username')),
+        #             course_id=row.get('course_id'),
+        #             role=row.get('role'),
+        #         ) for row in reader[i:i + batch_size]),
+        #         ignore_conflicts=True,
+        #     )
+        #     time.sleep(batch_delay)
--- a/edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
+++ b/edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
@@ -5,7 +5,7 @@ from django.core.management import call_command
 from django.test import TestCase
 
 from edx_exams.apps.core.models import CourseStaffRole, User
-from edx_exams.apps.core.test_utils.factories import UserFactory
+from edx_exams.apps.core.test_utils.factories import CourseStaffRoleFactory, UserFactory
 
 
 class TestBulkAddCourseStaff(TestCase):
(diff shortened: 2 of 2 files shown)


edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -84,11 +84,11 @@ def add_course_staff_from_csv(self, csv_file, batch_size, batch_delay):
...
             CourseStaffRole.objects.bulk_create(
-                CourseStaffRole(
+                (CourseStaffRole(
                     user=User.objects.get(username=row.get('username')),
                     course_id=row.get('course_id'),
                     role=row.get('role'),
-                )
-                for row in reader[i:i + batch_size]
+                ) for row in reader[i:i + batch_size]),
+                ignore_conflicts=True,
```
varshamenon4, line 92: Question: is it possible to just use `ignore_conflicts=True` here rather than taking the more convoluted approach above for users, see below for code snippet (where we first create a list of users that need to be created and don't exist, then use this for the bulk operation). 

Looking at the documentation, it seems like ignore_conflicts isn't completely supported for mysql: https://docs.djangoproject.com/en/5.0/ref/models/querysets/#bulk-create. But this line sugg

It is working when I test locally but this must be because we are using sqlite for local/test, which is supported. Oracle/MySQL does not support but I also see this. I am wondering if ignore_conflicts wasn't used before because of exactly this. It seems like this could have been the reason for taking the other approach in the first place but I just want to be sure.

> Warning

> On MySQL and MariaDB, setting the ignore_conflicts parameter to True turns certain types of errors, other than duplicate key, into warnings. Even with Strict Mode. For example: invalid values or non-nullable violations. See the [MySQL documentation](https://dev.mysql.com/doc/refman/en/sql-mode.html#ignore-strict-comparison) and [MariaDB documentation](https://mariadb.com/kb/en/ignore/) for more details.

```
users_to_create = []
        users_existing = {u.username for u in User.objects.filter(username__in=[r.get('username') for r in reader])}
        for row in reader:
            if row.get('username') not in users_existing:
                users_to_create.append(row)
                users_existing.add(row.get('username'))
        # bulk create users
        for i in range(0, len(users_to_create), batch_size):
            User.objects.bulk_create(
                User(
                    username=user.get('username'),
                    email=user.get('email'),
                )
                for user in users_to_create[i:i + batch_size]
            )
            time.sleep(batch_delay)
```

varshamenon4 (commented):
Question about ignore_conflicts

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: After speaking about this in standup, it looks like we can use ignore_conflicts. Tested locally and found other usages of ignore_conflicts within bulk_create in other places in the platform. So should be good to go here.

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -63,32 +63,26 @@ def add_course_staff_from_csv(self, csv_file, batch_size, batch_delay):
...
             User.objects.bulk_create(
-                User(
-                    username=user.get('username'),
-                    email=user.get('email'),
-                )
-                for user in users_to_create[i:i + batch_size]
+                (User(
+                    username=row.get('username'),
+                    email=row.get('email'),
+                ) for row in reader[i:i + batch_size]),
```
schenedx, line 73: Will this line generate indexing errors? For example, the `i+batch_size > len(reader)` condition?

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
```diff
@@ -130,3 +130,32 @@ def test_dupe_user_csv(self):
             call_command(self.command, f'--csv_path={csv.name}')
             self._assert_user_and_role(username, email, self.course_role, self.course_id)
             self._assert_user_and_role(username, email, self.course_role, course_id_2)
+
```
schenedx, line 133: Could you add a test case where you set the batch size to 2 and create like 3 course_staff users?

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: Good call, let me add testing/handling for that.

varshamenon4: So I did add tests to test that this still works. Also, I believe that this works because of this line: `for i in range(0, len(reader), batch_size)`. Since this is incremented by batch_size, the line `for row in reader[i:i + batch_size])` won't actually get to the index error. I actually originally repurposed this from some batching examples in the platform, so there are other examples of this pattern being used. See below:

- https://github.com/search?q=repo%3Aopenedx%2Fedx-platform%20%2B%20batch_size&type=code
- https://github.com/openedx/edx-platform/blob/8d0ada54c428b0c546119338d3661980ed4d23d2/scripts/user_retirement/utils/utils.py#L22
- https://github.com/openedx/edx-platform/blob/8d0ada54c428b0c546119338d3661980ed4d23d2/openedx/core/djangoapps/notifications/utils.py#L56

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
```diff
@@ -90,9 +90,18 @@ def test_add_course_staff_with_not_default_batch_size(self):
                  'sam,sam@pond.com,staff,course-v1:edx+test+f20\n']
         with NamedTemporaryFile() as csv:
             csv = self._write_test_csv(csv, lines)
-            with self.assertNumQueries(9):
+            with self.assertNumQueries(8):
                 call_command(self.command, f'--csv_path={csv.name}', '--batch_size=1')
 
+    def test_add_course_staff_with_batch_size_larger_than_list(self):
```
varshamenon4, line 96: @schenedx added this test case as you'd suggested! Thanks for the feedback.

edx_exams/apps/core/management/commands/test/test_bulk_add_course_staff.py
varshamenon4: Adding two cases, one where the batch size is larger than the number of roles, and one where the batch size is smaller.

varshamenon4 (commented):
Added some test cases!

schenedx (approved):
👍

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
Zacharis278: okay this was throwing me off too. `reader[i:i + batch_size]` is actually slicing a subset `reader` into a new list. If the `k` in list[i:k] is outside the bounds it won't matter it only pulls the subset of indexes that actually exist up to `k`.

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -63,32 +63,26 @@ def add_course_staff_from_csv(self, csv_file, batch_size, batch_delay):
...
+                ignore_conflicts=True,
             )
             time.sleep(batch_delay)
 
         # bulk create course staff
         for i in range(0, len(reader), batch_size):
```
Zacharis278, line 79: NIT: Is there any reason we need to do this in two separate loops? Works either way so I'd leave it up to you on if you think consolidating makes sense or no.

Zacharis278 (approved):
small nit but otherwise LGTM

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: You're right. I had two separate loops when we were creating an "existing_users" list to iterate on, but now that we're using ignore_conflics and iterating on the same list, I can remove before merging.

PR #308, opened 2024-08-21
feat: non-required fields optional in admin

Zacharis278:
I've had problems with this in the past doing things like making a 'created' attempt in admin which shouldn't have a start/end date but Django requires it without an explicit `blank=True`. Came across it again trying to setup LTI providers for partners where we only want a support chat url and not a phone number but both are 'required' according to admin form logic.

Went through and fixed the fields that made sense to be optional. This only impacts the admin forms and not the DB (these are already `null=True`)


varshamenon4 (approved):
lgtm!

PR #310, opened 2024-08-23
fix: course staff mgmt command

varshamenon4:
**JIRA:** [COSMO-234](https://2u-internal.atlassian.net/browse/COSMO-234)

**Description:** Attempting to fix the course staff role mgmt command by separating the loops and removing the atomic transaction. I am unable to replicate locally, but running this command in stage resulted in an error where the User couldn't be found.

### DIFF
--- a/edx_exams/apps/core/management/commands/bulk_add_course_staff.py
+++ b/edx_exams/apps/core/management/commands/bulk_add_course_staff.py
@@ -4,7 +4,7 @@ import logging
 import time
 
 from django.core.management.base import BaseCommand
-from django.db import transaction
+from django.db import IntegrityError, transaction
 
 from edx_exams.apps.core.models import CourseStaffRole, User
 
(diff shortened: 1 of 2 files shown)


edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -63,34 +63,19 @@ def add_course_staff_from_csv(self, csv_file, batch_size, batch_delay):
...
-            )
-            CourseStaffRole.objects.bulk_create(
-                (CourseStaffRole(
-                    user=User.objects.get(username=row.get('username')),
-                    course_id=row.get('course_id'),
-                    role=row.get('role'),
-                ) for row in reader[i:i + batch_size]),
-                ignore_conflicts=True,
-            )
+            users_dict = {k: v for k, v in [(u.username, u) for (u, c) in [User.objects.get_or_create(username=row.get('username'), email=row.get('email')) for row in reader[i:i + batch_size]]]}
```
varshamenon4, line 69: Does this look good? I can now use this to get users in the bulk create function as this does include the pk. Need to handle IntegrityError (specifically ran into the error when trying to add a user with an existing username, but a different email). Guessing this can be done with a try/catch?

varshamenon4 (commented):
Ready for review. Had one question about error handling included in comments!

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
varshamenon4: My question specifically related to error handling: what do we want to happen in this case? Ideally we'd want to skip the failing entry, right? So maybe the move is to break out the list comprehension (for readability sake) and to handle the error at the get_or_create by having a try/catch?

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
```diff
@@ -63,34 +63,25 @@ def add_course_staff_from_csv(self, csv_file, batch_size, batch_delay):
...
-                    course_id=row.get('course_id'),
-                    role=row.get('role'),
-                ) for row in reader[i:i + batch_size]),
-                ignore_conflicts=True,
-            )
+            users_list = []
+            for row in reader[i:i + batch_size]:
+                try:
+                    users_list.append(User.objects.get_or_create(username=row.get('username'), email=row.get('email')))
+                except IntegrityError:
```
varshamenon4, line 73: Took a pass at handling IntegrityError here. This handles the case where there's a duplicate username with a different email. This does not handle when there is a new username with a duplicate email address. I don't know if this is an issue, but this is also something that may need to be handled in the Model/db itself so might be more cumbersome than needed.

varshamenon4 (commented):
Took a pass at the handling of the error. But there still seems to be some corner cases that aren't working as I'd expect (see comment for more details). Not sure if this is relevant or needed for this work though.

edx_exams/apps/core/management/commands/bulk_add_course_staff.py
Zacharis278: if this is a case we're expecting can we log something or output information about the rows that failed to create when this is run so the failure isn't silent?

varshamenon4: Yes I can do that.

Zacharis278 (approved):

PR #359, opened 2025-02-13
fix: grant course staff exam access tokens

ilee2u:
**JIRA:** [Link to JIRA ticket](https://2u-internal.atlassian.net/browse/CR-7304)

**Description:** Add a fix to allow course staff to view the exams they create, as currently some courses staff are blocked from viewing the exams they've created.

migrations, etc.

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

### DIFF
--- a/edx_exams/apps/api/v1/tests/test_views.py
+++ b/edx_exams/apps/api/v1/tests/test_views.py
@@ -654,6 +654,8 @@ class ExamAccessTokensViewsTests(ExamsAPITestCase):
 
         self.course_id = 'course-v1:edx+test+f19'
 
+        self.user.is_staff = False
+
         self.due_date = timezone.now() + timedelta(minutes=5)
         self.exam = Exam.objects.create(
             resource_id=str(uuid.uuid4()),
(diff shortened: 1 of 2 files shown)


edx_exams/apps/api/v1/tests/test_views.py
```diff
@@ -815,6 +815,26 @@ def test_access_no_due_date(self, attempt_status, response_status):
         if response_status == 200:
             self.assert_valid_exam_access_token(response, self.user, no_due_date_exam)
 
+    def test_access_user_is_course_staff(self):
+        """
+        Verify the endpoint grants access for an exam
```
MichaelRoytman, line 820: If you agree with my suggestion below, please update the docstring and the test.

edx_exams/apps/api/v1/views.py
```diff
@@ -385,6 +385,10 @@ def get_response(cls, exam, user):
         elif exam.due_date is not None and timezone.now() >= exam.due_date:
             grant_access, response_status = True, status.HTTP_200_OK
 
+        # If user is course staff, then grant them access
```
MichaelRoytman, line 388: This is what we want, but I think we should have this before line 369. On the frontend, [access to the contents of the exam will always be granted to staff members](https://github.com/openedx/frontend-lib-special-exams/blob/4878aef9188a4b5df42ff0a2229b5d34a3624d09/src/exam/ExamWrapper.jsx#L45-L50). Because of this, they never see any exam functionality and will not be able to start an exam attempt. Therefore, we can short circuit early and give them access immediately if they are a staff member - no need to look at exam attempts. This makes the code easier to reason about. Does that match your understanding?

MichaelRoytman (commented):
Nice job! I left a comment.

MichaelRoytman (approved):
Nice!

